PR #1, opened 2016-07-01
Review for INFRA-32

jdswinbank:


bin.src/createDetGeom.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
+#createRefSpec.py '/Users/azuri/spectra/pfs/PFS' --id visit=4 filter='r' spectrograph=2 site='F' category='A' -C ../obs_pfs/config/pfs/createRefSpec.py --output "/Users/azuri/spectra/pfs/PFS" --clobber-config
```
jdswinbank: Comment here seems to refer to a different file.

bin.src/createDetGeom.py
```diff
@@ -0,0 +1,5 @@
...
+from pfs.drp.stella.createDetGeom import CreateDetGeomTask
+task = CreateDetGeomTask()
+task.run()
```
jdswinbank: I see from the discussion on HipChat what happened here: this is a "command line task" without being a `CmdLineTask`, and hence the weird set up. I agree with the commentary from @RobertLuptonTheGood on HipChat: this probably more convoluted than it needs to be, but it's not worth spending effort in refactoring so long as this does the job.

I do think it would be useful to add a comment describing what's going on, though, either to this file or to `createDetGeom.py` itself.

bin.src/createRefSpec.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env python
+#createRefSpec.py '/Users/azuri/spectra/pfs/PFS' --id visit=4 filter='r' spectrograph=2 site='F' category='A' -C ../obs_pfs/config/pfs/createRefSpec.py --output "/Users/azuri/spectra/pfs/PFS" --clobber-config
```
jdswinbank: This comment does seem to apply to this file, but it could do with a little context. Is this just an example of running the task that you might want to copy and paste? Is the content of the config override important?

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
+import os
+import sys
+import argparse
```
jdswinbank: Neither `sys` not `argparse` are used.

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
...
+class CreateDetGeomConfig(pexConfig.Config):
+    """Configuration for creating the PFS detector geometry files"""
+
+class CreateDetGeomTask(Task):
+    """Task to create the detector geometry files for PFS"""
+    ConfigClass = CreateDetGeomConfig
+    _DefaultName = "createDetGeomTask"
+
+    def __init__(self, *args, **kwargs):
+        super(CreateDetGeomTask, self).__init__(*args, **kwargs)
```
jdswinbank: LSST discourages the use of `super` unless it's really, really necessary: see [the Developer Guide](https://developer.lsst.io/coding/python_style_guide.html#super-should-not-be-used-unless-the-author-really-understands-the-implications-e-g-in-a-well-understood-multiple-inheritance-hierarchy). I don't know if PFS has formally adopted the LSST standards, but probably following them would be a good idea.

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def run(self):
+        gain = (1.24, 1.24, 1.27, 1.18, 1.26, 1.20, 1.24, 1.26)
+        rdnoise = (3.61, 3.78, 3.18, 2.95, 3.19, 3.80, 4.51, 3.18)
```
jdswinbank: It'd be nice to document where these come from. Maybe they're obvious to the veterans...

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
...
+        
+        outDir = os.getenv('OBS_PFS_DIR', '~/stella_git/obs_pfs')
+        print 'outDir = <',outDir,'>'
+        outDir = outDir+'/pfs/camera'
```
jdswinbank: At one point, this was configurable (albeit with an arguably unfortunate default). Now it's that. Why the change?

Actually, that makes me question something more fundamental: while I've not yet fully understood the structure of the PFS codebase, this script seems like something that relates directly to the PFS camera. Why doesn't it live in obs_pfs rather than drp_stella?

Finally, `os.path.join()` rather than string concatenation, and `lsst.utils.base.getPackageDir()` rather than `os.getenv()`.

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
...
+        
+        amp = 0
+        for iArm in range(3):
+            for iCCD in range(4):
+                detector = afwTable.AmpInfoCatalog(afwTable.AmpInfoTable.makeMinimalSchema())
+
+                for iAmp in range(8):
+
+                    x0 = iAmp * 552
+                    x1 = ((iAmp + 1) * 552)-1
```
jdswinbank: There's a lot of numbers hard-coded in here. Is that necessary or desirable? Could they change in future? Should they be configurable, or read from some separate data file, rather than directly embedded in Python? Can you provide references to the source material?

python/pfs/drp/stella/createDetGeom.py
```diff
@@ -0,0 +1,108 @@
...
+
+                sArm = ' '
+                if iArm == 0:
+                    sArm = 'b'
+                elif iArm == 1:
+                    sArm = 'r'
+                elif iArm == 2:
+                    sArm = 'n'
+                else: 
+                    sArm = 'm'
```
jdswinbank: You could replace the clunky `if` block with something more like this, I think:

    sArm = "brnm"[iArm]

python/pfs/drp/stella/createRefSpec.py
```diff
@@ -0,0 +1,118 @@
...
+import pfs.drp.stella as drpStella
+import lsst.daf.persistence as dafPersist
+import lsst.afw.image as afwImage
+from lsst.pex.config import Config, ConfigField, ConfigurableField, Field, ListField
+from lsst.pipe.base import Task, Struct, TaskRunner, ArgumentParser, CmdLineTask
+import lsst.daf.persistence.butler as lsstButler
+import numpy as np
+from astropy.io import fits as pyfits
+import matplotlib.pyplot as plt
+
```
jdswinbank: All of `os`, `sys`, `argparse`, `dafPersist`, `afwImage,` `ConfigField`, `ListField`, `ConfigurableField`, `Task`, `lsstButler`, `plt` are unused!

python/pfs/drp/stella/createRefSpec.py
```diff
@@ -0,0 +1,118 @@
...
+                               help="input identifiers, e.g., --id visit=123 ccd=4")
+        parser.add_argument("--lineList", help='directory and name of line list')
+        return parser
+
+    def run(self, expRefList, butler, lineList=None, immediate=True):
+        print 'expRefList = ',expRefList
+        print 'expRefList[0] = ',expRefList[0]
+        arcRef = expRefList[0]
+        try:
+            arcExp = arcRef.get("postISRCCD", immediate=True)
```
jdswinbank: `immediate=True` or `immediate=immediate`?

python/pfs/drp/stella/createRefSpec.py
```diff
@@ -0,0 +1,118 @@
...
+            raise RuntimeError("Unable to retrieve %s: %s" % (arcRef.dataId, e))
+
+	try:
+            flatExposure = arcRef.get('flat', immediate=immediate)
+        except Exception, e:
+            raise RuntimeError("Unable to retrieve flat for %s: %s" % (arcRef.dataId, e))
+
+        """ find and trace fiber traces """
+        print 'tracing flat fiber traces'
+        myFindTask = fataTask.FindAndTraceAperturesTask()
```
jdswinbank: As a total neophyte here, I find the naming a bit confusing. Find and trace fibre traces? Or *flat* fibre traces? Or apertures? Are they all the same? (I assume the answer is yes, but could we at least use the same terminology on the three consecutive lines?)

python/pfs/drp/stella/createRefSpec.py
```diff
@@ -0,0 +1,118 @@
...
+        print 'calculating spatial profiles'
+        myProfileTask = cfftpTask.CreateFlatFiberTraceProfileTask()
+        myProfileTask.run(flatFiberTraceSet)
+        
+        """ optimally extract arc spectra """
+        print 'extracting arc spectra'
+
+        """Extract all apertures"""
+        myExtractTask = esTask.ExtractSpectraTask()
+        aperturesToExtract = [-1]
```
jdswinbank: Again, without knowing the codebase more intimately than I do I can't be sure, but `aperturesToExtract = [-1]` doesn't *obviously* agree with the comment that we're extracting all apertures.

python/pfs/drp/stella/createDetGeom.py
ziggyman: Comment added: """ Gain and RdNoise from Jim Gunn's talk in Marseille Dec 2015"""

bin.src/createDetGeom.py
ziggyman: changed to an example call of createDetGeom.py, added "# example:" before

bin.src/createDetGeom.py
ziggyman: Added a comment to bin.src/createDetGeom.py stating that CreateDetGeomTask is a Task and not a CmdLineTask

bin.src/createRefSpec.py
ziggyman: Added "# example:" before "#createRefSpec..." and removed the config override for the example, instead added the two config parameters as arguments

python/pfs/drp/stella/createDetGeom.py
ziggyman: removed sys and argparse

python/pfs/drp/stella/createDetGeom.py
ziggyman: replaced 
"super(CreateDetGeomTask, self).__init__(*args, **kwargs)"
with
"Task.__init__(self, *args, **kwargs)",
similar in createRefSpecTask

python/pfs/drp/stella/createDetGeom.py
ziggyman: The output directory will not change so I removed it from the parameter list.
You are right, this should be an obs_pfs task. Originally it was an ipython notebook in drp_stella, before obs_pfs was created. Now it should really go into obs_pfs, so I moved both createDetGeom.py (bin.src and python/pfs/drp/stella) to obs_pfs (bin.src and python/lsst/obs/pfs)

python/pfs/drp/stella/createRefSpec.py
ziggyman: If a number >= 0 is given then the aperture with that number will get extracted. -1 marks all apertures.

python/pfs/drp/stella/createRefSpec.py
ziggyman: changed "fiber traces" to "apertures"

python/pfs/drp/stella/createRefSpec.py
ziggyman: changed "immediate=True" to "immediate=immediate"

python/pfs/drp/stella/createRefSpec.py
ziggyman: removed unused imports

python/pfs/drp/stella/createDetGeom.py
ziggyman: done

PR #3, opened 2016-08-05
Review for INFRA-36: Quick start guide

jdswinbank:


doc/install.text
```diff
@@ -0,0 +1,124 @@
+The following notes should allow you to install the LSST stack and the PFS DRP.
+After the installation procedure example commands are given to show you how to
+use the pipeline. The commands have been tested on an Arch Linux machine as well
+as on Mac OS X.
```
jdswinbank: Is there a reference platform for PFS? e.g. LSST guarantees everything will work on CentOS 7.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+* Install lsst binary distribution (https://pipelines.lsst.io/install/conda.html)
```
jdswinbank: Does it matter if you install Anaconda or Miniconda?

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+* Install git-lfs (https://git-lfs.github.com/)
+* Install matplotlib:
+
+conda install matplotlib
+
+
+If you have a very fresh compiler version, this could lead to incompatibilities 
+with the precompiled LSST stack. In case you run into trouble compiling or testing
+the LSST stack please install your compiler in anaconda:
```
jdswinbank: "Recent" rather than "fresh"?

Can we be more specific about exactly what versions of the compiler are required, and what the error is?

Alternatively: is there any downside to simply requiring that the user always install gcc through Conda? I don't think this can do any harm, and avoids a class of problems.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+conda install gcc
+
+
+* Create a directory for the PFS-DRP repositories:
+
+mkdir ~/stella-git
+cd ~/stella-git
```
jdswinbank: We probably don't want to be prescriptive about storing this in the reader's home directory.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+* Clone the relevant git repositories:
+
+git clone https://github.com/Subaru-PFS/drp_stella.git
+git clone https://github.com/Subaru-PFS/drp_stella_data.git
+git clone https://github.com/Subaru-PFS/obs_pfs.git
+
+
+* Install a newer version of eigen:
+
+eups distrib install eigen 3.2.5.lsst2
```
jdswinbank: This gives me:

```
$ eups distrib install eigen 3.2.5.lsst2
eups distrib install: No repositories specified; please set -r or $EUPS_PKGROOT
```

(`export EUPS_PKGROOT="https://sw.lsstcorp.org/eupspkg/"` fixed it, of course)

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* Declare the git repositories in Linux (replace <Linux64> with <DarwinX86> for Mac OS X):
+
+eups declare obs_pfs 1.0 -c -f Linux64 -r ~/stella-git/obs_pfs
+eups declare drp_stella_data 1.0 -c -f Linux64 -r ~/stella-git/drp_stella_data
+eups declare drp_stella 1.0 -c -f Linux64 -r ~/stella-git/drp_stella
```
jdswinbank: Do you need to specify the architecture? Seems to work fine without for me.

Do you want to declare these as version 1.0? That sounds like a finished product. I don't think there is a usable versioning scheme, but you could make something up from the date (`echo 0.0.$(date +%Y%m%d)`) or following PEP-440 or something. Or just make them 0.1 instead!

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* Setup the pipeline. As obs_pfs and drp_stella_data are required packages they are set up automatically:
+
+setup drp_stella 1.0 -v
```
jdswinbank: This says (lots of stuff, followed by):

```
Setting up: |eigen                          Flavor: Linux64    Version: 3.2.5.lsst2
             eigen 3.2.5.lsst1 is currently setup; overriding with 3.2.5.lsst2
Setting up: |sconsUtils                     Flavor: Linux64    Version: 12.0
Setting up: |drp_stella_data                Flavor: Linux64    Version: 1.0
             ... optional setup drp_stella_data requested by drp_stella 1.0 failed
```
Issue seems to be the hard-coded `sconsUtils` version in drp_stella_data:
```
$ setup drp_stella_data
setup: in file /ssd/swinbank/pfs/stella-git/drp_stella_data/ups/drp_stella_data.table: Product sconsUtils 12.0-1-g08b90fd not found
```
I took that out (by editing the table file) in order to continue with the instructions.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+should show you:
+
+~/stella-git/drp_stella
+
+
+* Build the pipeline:
+
+cd drp_stella
+scons opt=3 -j8
```
jdswinbank: Why `-j8`? Seems unambitious on this machine:
```
$ cat /proc/cpuinfo  | grep processor | tail -1
processor	: 47
```

:-)

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* Create the defects database (done during compilation of the code):
```
jdswinbank: So is our main aim to compile the code or create the defects database? The phrasing seems a bit awkward.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+cd ../obs_pfs
+scons opt=3 -j8
+
+
+* Now for using the pipeline. Raw test data are in $DRP_STELLA_DATA_DIR/tests/data/raw/ (3 Biases, 3 Darks, 1 Flat, 1 Arc). 
+First we need to create a directory (actually 2) where we want to store pipeline outputs. 
+Let's assume you have a directory ~/spectra and do everything related to spectra there:
+
+mkdir ~/spectra/PFS
+mkdir ~/spectra/PFS/CALIB
```
jdswinbank: You can do this with just one command: `mkdir -p ~/spectra/PFS/CALIB`. That also means we can cut the quibbling about whether we're making one directory or two.

Rather than assuming we're working in the home directory, might be nice to be less explicit. But that's not a big deal.

doc/install.text
jdswinbank: Note that drp_stella on the `tickets/INFRA-36` branch doesn't build:
```
In file included from include/pfs/drp/stella/math/CurveFitting.h:12:0,
                 from src/math/CurveFitting.cc:1:
include/pfs/drp/stella/math/Math.h:17:79: fatal error: /Users/azuri/Downloads/eigen/test/i
nclude/unsupported/Eigen/Splines: No such file or directory
 #include "/Users/azuri/Downloads/eigen/test/include/unsupported/Eigen/Splines"
```
This problem doesn't seem to exist on `master`.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* Configuration parameters can be passed to the pipeline task either on the command line or in a config file 
+(e.g. $OBS_PFS_DIR/config/pfs/ingest.py, see following example). You can list all possible configuration parameters by 
+appending a "--show config" to the parameter list.
+To ingest the raw images:
+
+cd ~/spectra
+ingestImages.py PFS $DRP_STELLA_DATA_DIR/tests/data/raw/*.fits --output=PFS --mode=link -C $OBS_PFS_DIR/config/pfs/ingest.py
```
jdswinbank: Fails:
```
$ ingestImages.py PFS $DRP_STELLA_DATA_DIR/tests/data/raw/*.fits --output=PFS --mode=link -C $OBS_PFS_DIR/config/pfs/ingest.py
Traceback (most recent call last):
  File "/ssd/swinbank/pfs/miniconda2/envs/lsst/opt/lsst/pipe_tasks/bin/ingestImages.py", line 3, in <module>
    IngestTask.parseAndRun()
  File "/ssd/swinbank/pfs/miniconda2/envs/lsst/opt/lsst/pipe_tasks/python/lsst/pipe/tasks/ingest.py", line 364, in parseAndRun
    args = parser.parse_args(config)
  File "/ssd/swinbank/pfs/miniconda2/envs/lsst/opt/lsst/pipe_base/python/lsst/pipe/base/argumentParser.py", line 430, in parse_args
    mapperClass = dafPersist.Butler.getMapperClass(namespace.input)
  File "/ssd/swinbank/pfs/miniconda2/envs/lsst/opt/lsst/daf_persistence/python/lsst/daf/persistence/butler.py", line 185, in getMapperClass
    return PosixStorage.getMapperClass(root)
  File "/ssd/swinbank/pfs/miniconda2/envs/lsst/opt/lsst/daf_persistence/python/lsst/daf/persistence/posixStorage.py", line 107, in getMapperClass
    (mapperFile,))
RuntimeError: No mapper provided and no _mapper available
```
Fixed by:
```
$ cat > PFS/_mapper
lsst.obs.pfs.PfsMapper
^D
```

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* You can now inspect the created SQL database with e.g. sqlitebrowser (http://sqlitebrowser.org/).
+Now that we have our database we can start reducing things. We should probably start with the Biases. If the biases we
+want to reduce were observed on 2015-12-22 on spectrograph 2, arm red (r) at site S:
```
jdswinbank: Were they? We're using the test data in drp_stella_data, so you should know!

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+* Configuration parameters can be passed to the pipeline task either on the command line or in a config file 
+(e.g. $OBS_PFS_DIR/config/pfs/ingest.py, see following example). You can list all possible configuration parameters by 
+appending a "--show config" to the parameter list.
+To ingest the raw images:
+
+cd ~/spectra
+ingestImages.py PFS $DRP_STELLA_DATA_DIR/tests/data/raw/*.fits --output=PFS --mode=link -C $OBS_PFS_DIR/config/pfs/ingest.py
+
+
+* You can now inspect the created SQL database with e.g. sqlitebrowser (http://sqlitebrowser.org/).
```
jdswinbank: ...or by just running `sqlite3`, which is more likely to be available. But what are we expecting the reader to gain by doing so? It seems like it'll just invite scratching-of-heads.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+Now that we have our database we can start reducing things. We should probably start with the Biases. If the biases we
+want to reduce were observed on 2015-12-22 on spectrograph 2, arm red (r) at site S:
+
+reduceBias.py PFS --calib PFS/CALIB --output PFS --calibId calibVersion=bias arm=r calibDate=2015-12-22 spectrograph=2 --do-exec --id field=BIAS dateObs=2015-12-22 --clobber-config --nodes=1 --procs=1
```
jdswinbank: This produces lots of output that I didn't expect (list of set up packages, a dump of the environment, etc etc). I guess that's because it's a `BatchPoolTask`? If so, I guess that's an LSST quirk.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+
+* Note the 2 config parameters --nodes and --procs at the end. These parameters are required by tasks which are 
+parallelized. Sometimes running the code in parallel can lead to problems (in most cases caused by the 3rd-party libraries
+used), so setting nodes and procs to 1 is a safe choice. —-clobber-config is needed if you re-run the task with different
+config parameters. The config used for each run is written to PFS/config/bias.py and the task won’t run again if you 
+change parameters, unless you say —-clobber-config.
```
jdswinbank: I'd suggest not using `--clobber-config` here: it's really only something users would want in exceptional circumstances.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+* Now that we have a master bias we need to inject that too into our database:
+
+genCalibRegistry.py --root=PFS/CALIB --camera PFS --validity 180
+
+* Now we can create a bias-subtracted Dark and ingest that into our database:
+
+reduceDark.py PFS --calib PFS/CALIB --output PFS --calibId calibVersion=dark arm=r calibDate=2015-12-22 spectrograph=2 --do-exec --id field=DARK dateObs=2015-12-22 --nodes=1 --procs=1 -C ~/stella-git/obs_pfs/config/pfs/dark.py
```
jdswinbank: Why do we need to specify a config file here? We didn't for the biases above.

Besides that: this config will get picked up by default anyway. No need to mention it on the command line.

doc/install.text
jdswinbank: No need to specify the config override on the command line here — it'll be picked up automatically.

doc/install.text
jdswinbank: This fails:
```
dark WARNING: Unable to process {'category': 'A', 'taiObs': '2015-12-22', 'visit': 7291, 'dateObs': '2015-12-22', 'site': 'S', 'filter': 'n', 'field': 'DARK', 'spectrograph': 1, 'ccd': 8, 'arm': 'n'}: Una
ble to retrieve bias for {'category': 'A', 'taiObs': '2015-12-22', 'visit': 7291, 'dateObs': '2015-12-22', 'site': 'S', 'filter': 'n', 'field': 'DARK', 'spectrograph': 1, 'ccd': 8, 'arm': 'n'}: No unique
lookup for ['calibDate', 'calibVersion'] from {'category': 'A', 'taiObs': '2015-12-22', 'visit': 7291, 'dateObs': '2015-12-22', 'site': 'S', 'filter': 'n', 'field': 'DARK', 'spectrograph': 1, 'ccd': 8, 'a
rm': 'n'}: 0 matches
```
My understanding is that the data you've provided in drp_stella_data us for spectrograph 1, arm n; however, your commands in this text keep referring to spectrograph 2, arm r. By re-running `reduceBias.py` with an appropriate `calibId` I was able to eventually get `reduceDark.py` to run.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+genCalibRegistry.py --root=PFS/CALIB --camera PFS --validity 180
+
+* and a bias and dark-subtracted flat (currently only used to trace the apertures of the fiber traces). Note that in this
+example we set the config parameters isr.doBias and isr.doDark to True after <—-config > (instead of the parameter file
+after <-C > in the ingestImages.py example above):
```
jdswinbank: Why this inconsistency in configuration?

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+reduceFlat.py PFS --calib PFS/CALIB --calibId calibVersion=flat arm=r calibDate=2015-12-22 filter=r spectrograph=2 --do-exec --id field=FLAT arm=r dateObs=2015-12-22 spectrograph=2 --loglevel info --config isr.doBias=True isr.doDark=True isr.doLinearize=False --output PFS --nodes=1 --procs=1 --clobber-config
+
+genCalibRegistry.py --root=PFS/CALIB --camera PFS --validity 180
+
+* Since we have the Bias and Dark we can also run the Instrumental-Signature Removal (ISR) task for our Arc spectrum:
+
+detrend.py PFS --calib=PFS/CALIB --id arm=r spectrograph=2 dateObs=2015-12-22 field=ARC -C ~/stella-git/obs_pfs/config/pfs/detrend.py --output PFS
```
jdswinbank: Again (as in all the above), I needed to use arm n, spectrograph 1.

doc/install.text
```diff
@@ -0,0 +1,124 @@
...
+
+* We now have the postISRCCD images and can extract and wavelength-calibrate our CdHgKrNeXe Arc with the visit
+number 4:
+
+reduceArc.py PFS --id visit=4 --wLenFile $OBS_PFS_DIR/pfs/RedFiberPixels.fits.gz --lineList $OBS_PFS_DIR/pfs/lineLists/CdHgKrNeXe_red.fits --loglevel info --calib PFS/CALIB/ --output PFS
```
jdswinbank: Why do we need to adjust log levels here?

jdswinbank: This ultimately finishes by saying:
```
reduceArcTask FATAL: Failed: [Errno 2] No such file or directory: '/Users/azuri/entwicklung/tex/reports/PFS_pipeline_status_2016-03/images/flux_vs_pix.pdf'
```
Which is pretty sad!

doc/install.text
ziggyman: Not that I know of...

doc/install.text
ziggyman: Honestly I don't know, but I believe that it shouldn't make a difference. Where did you setup that ubuntu machine where you tested the gcc (new version) bug? I could try to install miniconda there and try it out...

jdswinbank: I used a VM from https://www.digitalocean.com, but they cost real money (albeit only a few cents an hour), so I don't recommend that being a regular modus operandi.

For this purpose —  just testing whether we need Miniconda or Anaconda — I don't think we need anything special on the machine: any of the systems we have available should do the trick.

In fact, when I followed your instructions I used Miniconda and they worked fine (modulo my comments below); since Anaconda is a superset of that, I think we can conclude that either would be ok.

doc/install.text
ziggyman: Removed the paragraph and added gcc to "conda install matplotlib gcc"

doc/install.text
ziggyman: Added a line 
export EUPS_PKGROOT="https://sw.lsstcorp.org/eupspkg/"

doc/install.text
ziggyman: Removed the -f flag (guess that was a leftover from having a mix-up of lsst and lsstsw stacks on my laptop). Changed version number to 0.1.

doc/install.text
ziggyman: rebased drp_stella tickets/INFRA-36 to master

doc/install.text
ziggyman: removed the line "* Create...", running scons for obs_pfs is now part of "* Build the pipeline"

doc/install.text
ziggyman: Added instructions on how to create the _mapper file, removed config override on the command line

doc/install.text
ziggyman: Added an environment variable which the user can set to any directory they want

doc/install.text
ziggyman: Added environment variable SPECTRA_DIR and changed to mkdir -p $SPECTRA_DIR/CALIB

doc/install.text
ziggyman: With the fixed filename interpretation the arm changed to 'b'. Reformulated the sentence.

doc/install.text
ziggyman: Removed this step

doc/install.text
ziggyman: That's an LSST quirk...

doc/install.text
ziggyman: Removed --clobber-config

doc/install.text
CraigLoomis: There is a gotcha here.

We have been investigating simulated spots for the NIR arm; those are the most challenging, and the ones we worry most about understanding.

But the simulated & real images that we currently have are for/from CCDs. Specifically one of the red ones. Since a) we had real red CCD images and b) we understand the ISR components for HSC CCDs it made sense to simulate the NIR spots onto those CCD "detectors".

Yes we will have an ISR for the IR ramps, but we don't have one yet. So on the command line <arm=n> is probably OK, but the raw images really are quite different from what the <arm=n> ramp files will look like.

I think that in the end we will want to remove all traces of this confusing little fib. In the meanwhile it is probably better to use <arm=r> in as many places as possible.

doc/install.text
RobertLuptonTheGood: I don't think that's a good idea (it implies that you need to set it).  Especially as you give it a name ending _DIR which looks like a eups product

ziggyman: "Let's assume you want to store the pipeline outputs in a directory ~/spectra/PFS:

export SPECTRA_DIR="~/spectra/PFS"
mkdir -p $SPECTRA_DIR/CALIB"

suggestions?

doc/install.text
ziggyman: Removed the log level adjustment and disabled plots in reduceArcTask

doc/install.text
ziggyman: changed arm and spectrograph

doc/install.text
ziggyman: I wanted to show different ways of setting config parameters but I guess we just use the config files which are automatically read...

jdswinbank: This relates to [my comment on the ticket](https://pfspipe.ipmu.jp/jira/browse/INFRA-36?focusedCommentId=11225) — what's your audience for this guide? I'd suggest that what you want to provide is basic information about running drp_stella, not how to use the pipe_base framework, but perhaps that's something we could discuss if you disagree.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  git lfs install
+
+  There's no need to issue any "git lfs track" commands
+
+
+* On Linux, all the anaconda packages were compiled with gcc 4.small and are incompatible
+  with modern versions. Therefore, if you are installing the PFS DRP on a Linux machine, 
+  install gcc in anaconda:
+
+conda install gcc
```
jdswinbank: Have we checked what happens on a Mac?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+* For convenience of this quick-start guide, create a directory for the PFS-DRP repositories:
+
+export PFS_DRP=<your_target_directory>
+(e.g. export PFS_DRP=$HOME/PFS
```
jdswinbank: Missing ")".

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+* On Linux, all the anaconda packages were compiled with gcc 4.small and are incompatible
+  with modern versions. Therefore, if you are installing the PFS DRP on a Linux machine, 
+  install gcc in anaconda:
+
+conda install gcc
+
+
+* For convenience of this quick-start guide, create a directory for the PFS-DRP repositories:
```
jdswinbank: I think at least part of RHL's request here was not just that you explain that you wanted a working directory (which is arguably obvious) but that you're creating a particular environment variable to refer to it it — there's nothing special about the name ``$DRP_STELLA`` or ``$PFS_DRP`` except that it's convenient and you'll refer to it below. I think you could still make that clearer.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+export PFS_DRP=<your_target_directory>
+(e.g. export PFS_DRP=$HOME/PFS
+mkdir -p $PFS_DRP
+cd $PFS_DRP
+
+
+* Clone the relevant git repositories. Note that currently a newer version of sconsUtils
+  is needed to suppress some irrelevant compiler warnings. Installing and setting up
+  sconsUtils will be obsolete in the next binary LSST Stack installation.
```
jdswinbank: I like squashing compiler warnings in principle, but adding a custom sconsUtils makes things more complicated, and I'm guessing that folks who need to follow these instructions are not going to be carefully checking for build issues. Is this really necessary?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+git clone https://github.com/Subaru-PFS/drp_stella.git
+git clone https://github.com/Subaru-PFS/drp_stella_data.git
+git clone https://github.com/Subaru-PFS/obs_pfs.git
+git clone git@github.com:lsst/sconsUtils.git
+
+* The next 2 steps are temporary workarounds until fixed:
+* Set $EUPS_PKGROOT
```
jdswinbank: Are these temporary workarounds any more temporary than the sconsUtils material above? I'm not sure why you call them out separately.

jdswinbank: (Ok, the answer is because Robert told you to, but I think basically the same logic applies: all of this should become obsolete with the next binary release.)

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+* Install Eigen with the unsupported modules:
+
+eups distrib install eigen 3.2.5.lsst2
+
+
+* Build and setup the pipeline. Note that setting up drp_stella will overwrite
+  the already set up sconsUtils version with the standard one, so we need to
+  setup sconsUtils again before compiling drp_stella. There are a few warnings
```
jdswinbank: This seems complicated. Is it really necessary?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  when compiling drp_stella, one type of warnings is a doxygen warning "... not
+  found in argument list..." and the other is a SWIG problem ("...destination for
+  this 'memset' call is a pointer..."). Both types of warnings will be silenced 
+  in future version and can be safely ignored.
```
jdswinbank: Is this a response to an RHL complaint? Personally, I don't think it's appropriate for this document, but if he's told you to put it in then fair enough.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+
+
+
+
+
+
+
+
```
jdswinbank: That's a lot of empty lines. Why?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+setup -r .
+scons -Q opt=3 -j8 --filterWarn
+
+cd $PFS_DRP/drp_stella_data
+setup -r .
+
+cd $PFS_DRP/drp_stella
+setup -r .
+setup -r ../sconsUtils
+scons -Q opt=3 -j8 --filterWarn
```
jdswinbank: I understand you are using `setup -r` because Robert grumped about declaring things. However, if we're doing things this way, we need to explain how a user should set things up the next time they want to use this code. To wit: they will need to do the `setup`s, but not the `scons` commands (and obviously not the cloning, etc, above).

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  Raw test data are in $DRP_STELLA_DATA_DIR/tests/data/raw/:
+    - 3 Biases: visit numbers 7251-7253,
+    - 3 Darks: visit numbers 7291-7293,
+    - 1 Flat: visit number 5,
+    - 1 Arc: visit number 4 
+  Configuration parameters for the pipeline tasks can be set either in config files
+  (see $OBS_PFS_DIR/config/pfs/) or by passing them on the command line (after 
+  "--config", e.g. --config isr.doDark=False). You can list all configuration
+  parameters by appending a "--show config" to the parameter list. Setting "--show
+  config=glob"
```
jdswinbank: ... seems like there's something missing there.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+
+
+  First we need to create a directory (actually 2) where we want to store pipeline outputs. 
+  Let's assume you want to store the pipeline outputs in a directory $HOME/spectra/PFS.
+  For the convenience of this quick-start guide we define another environment variable:
+
+export PFS_DATA=$HOME/spectra/PFS
+mkdir -p $PFS_DATA/CALIB
```
jdswinbank: Could re-use the env var you set earlier, rather than a new one?

Also, explain why two directories are needed. (In fact, I'd drop the "(actually 2)" thing altogether: it's not useful.)

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+* We need to tell the LSST stack which mapper to use. The mapper gives a logical view
+  of both the raw data and the pipeline outputs which you can query for particular data
+  sets. It abstracts out the physical layout so we don't have to worry about the details.
```
jdswinbank: Can one query a view? Hmm.

How about: "The mapper provides a logical view of both the raw data and pipeline outputs, and provides facilities for querying for particular data sets. It abstracts away the details of the underlying storage, so we can avoid worrying about implementation details."?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+echo "lsst.obs.pfs.PfsMapper" > $PFS_DATA/_mapper
+
+
+* We can now ingest the raw images into a registry stored in $PFS_DATA/registry.sqlite3.
+  The "--mode link" parameter tells the pipeline to create symbolic links instead of copying
+  the raw images. The "-L warn" parameter sets the log level to only print warnings.
```
jdswinbank: Are you ingesting to a registry or a repository? You can't symlink into an sqlite database. (Of course I know what you mean here, but it's rather confusing.)

Also: is the `-L warn` stuff necessary? Is it suppressing warnings to keep Robert happy? I think I'd rather skip it for simplicity, unless it's *really* important.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+* Now that we have our database we can start reducing things. We start with creating a master Bias, followed
+  by a Bias-subtracted master Dark. We will then create a Bias- and Dark-subtracted master Flat, which we then use 
+  to identify and trace the apertures of the fiber traces. The fiber traces from the Arc image are then extracted
+  and wavelength calibrated.
+
+  The data we want to reduce were observed/simulated on 2015-12-22 on spectrograph 2, arm r (red) at site S (Summit). 
+
+  The parameter "--rerun USERNAME/rmp" (substitute USERNAME with your name) specifies where to store
+  temporary pipeline outputs.
```
jdswinbank: This is true, but I think it's confusing. Can we link to LSST documentation on the `--rerun` option? (Pretty sure pipe_base's Doxygen says something useful.)

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  Note the parameter "--cores 1" at the end. This parameter is required by tasks which are parallelized.
+  Sometimes running the code in parallel can lead to problems (in most cases caused by the 3rd-party 
+  libraries used), so setting cores to 1 is a safe choice.
```
jdswinbank: I find the idea that parallel execution is dangerous but must be explicitly disabled a little alarming! Is this really an issue? Would things break if we specified `--cores 2`?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+constructBias.py $PFS_DATA --rerun USERNAME/tmp --id field=BIAS dateObs=2015-12-22 arm=r spectrograph=2 --calibId calibVersion=bias calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
```
jdswinbank: I think it'd be helpful to break down the structure of this command in a little more detail. What's the distinction between `--id` and `--calibId`? Why do you need to specify so many things twice?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+
+* Now that we have a master bias we need to ingest that into our calibration database stored in
+  $PFS_DATA/CALIB/calibRegistry.sqlite3. The parameter "--validity 180" specifies that the calibration
+  images are valid for 180 days. We will need to repeat this step every time we create a new calibration
+  image for the butler to find it when its use is required by a later called task.
```
jdswinbank: What's a butler? You've not mentioned that before.

jdswinbank: (I don't think you should go into a big digression on the Butler here — probably just reword the above to avoid needing to explain it.)

doc/install.text
RobertLuptonTheGood: Yes, please provide a simple way for people to use this with the minimal number of magic commands.  Do you install the built package after building it?

doc/install.text
RobertLuptonTheGood: I'd like to call out that this will go away in the new release;  it's the only place that eups distrib appears.

doc/install.text
RobertLuptonTheGood: No need to make them do this.  It was for your benefit.

doc/install.text
RobertLuptonTheGood: No need to include this.  You need to understand the issues.

doc/install.text
RobertLuptonTheGood: Isn't this the directory for the data, and the previous one was where you built the code?

doc/install.text
RobertLuptonTheGood: Maybe remove it and add a line that "-L warn" makes the script much less verbose?  Our scripts are too verbose at default levels.

doc/install.text
RobertLuptonTheGood: id specified the inputs, calibId defines the outputs.  Please clarify this.  Is there a ticket to remove the calibVersion key (as we discussed with Paul)?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  in the config file:
+
+reduceFlat.py $PFS_DATA --rerun USERNAME/tmp --id visit=5 --calibId calibVersion=flat calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Since we have the Bias and Dark we can also perform the Instrumental-Signature Removal (ISR) task for our Arc
+  spectrum (visit=4). The ISR task detrend.py will subtract the Bias and scaled Dark from our Arc image. As
+  mentioned above, flat-fielding is not yet supported by the pipeline, but will be in the near future.
+  In case you would want to reduce all Arcs taken 2015-12-22 for spectrograph 2, red arm, simply replace
```
jdswinbank: (Again, no "would")

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  Again, if you don't want Dark subtraction, either edit the config file $OBS_PFS_DIR/config/pfs/detrend.py or add 
+  "--config isr.doDark=False" at the end of the command:
+
+detrend.py $PFS_DATA --output $PFS_DATA --id visit=4
+
+
+* We now have the postISRCCD images for our Flat and Arc and can extract and wavelength-calibrate our CdHgKrNeXe
+  Arc with the visit number 4:
+
+reduceArc.py $PFS_DATA --rerun USERNAME/tmp --id visit=4
```
jdswinbank: Will it be obvious to a user who has needed the level of instructions you've provided above how to work with the results when this runs? Maybe a few words of explanation would be appropriate?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  If you want to turn off Dark subtraction you can either edit the config file $OBS_PFS_DIR/config/pfs/flat.py
+  or simply add "--config isr.doDark=False" to the reduceFlat command, what will override the parameter setting
+  in the config file:
+
+reduceFlat.py $PFS_DATA --rerun USERNAME/tmp --id visit=5 --calibId calibVersion=flat calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Since we have the Bias and Dark we can also perform the Instrumental-Signature Removal (ISR) task for our Arc
+  spectrum (visit=4). The ISR task detrend.py will subtract the Bias and scaled Dark from our Arc image. As
```
jdswinbank: Being pedantic, the `.py` file isn't a task.

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  you would need to replace "visit=5" with "field=FLAT arm=r dateObs=2015-12-22 spectrograph=2".
+  If you want to turn off Dark subtraction you can either edit the config file $OBS_PFS_DIR/config/pfs/flat.py
+  or simply add "--config isr.doDark=False" to the reduceFlat command, what will override the parameter setting
+  in the config file:
+
+reduceFlat.py $PFS_DATA --rerun USERNAME/tmp --id visit=5 --calibId calibVersion=flat calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Since we have the Bias and Dark we can also perform the Instrumental-Signature Removal (ISR) task for our Arc
```
jdswinbank: "can also" -> "can now"?

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+* Having the master Bias and Dark, we can now create our master Flat. Currently the master Flat is only used to 
+  trace the apertures of the fiber traces and to calculate the spatial profile for the optimal extraction.
+  Note that for the actual flat-fielding dithered Flats will be used in the near future, and this kind of Flat
+  here will be renamed to apDef (aperture definition).
+  In our data set only visit 5 is a flat, so specifying "--id visit=5" is all you need to specify for the butler
+  to find our flat. In case you  would want to reduce all Flats taken 2015-12-22 for spectrograph 2, red arm,
+  you would need to replace "visit=5" with "field=FLAT arm=r dateObs=2015-12-22 spectrograph=2".
+  If you want to turn off Dark subtraction you can either edit the config file $OBS_PFS_DIR/config/pfs/flat.py
+  or simply add "--config isr.doDark=False" to the reduceFlat command, what will override the parameter setting
```
jdswinbank: "which will" rather than "what will".

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+constructDark.py $PFS_DATA --rerun USERNAME/tmp --id field=DARK dateObs=2015-12-22 arm=r spectrograph=2 --calibId calibVersion=dark calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Having the master Bias and Dark, we can now create our master Flat. Currently the master Flat is only used to 
+  trace the apertures of the fiber traces and to calculate the spatial profile for the optimal extraction.
+  Note that for the actual flat-fielding dithered Flats will be used in the near future, and this kind of Flat
+  here will be renamed to apDef (aperture definition).
+  In our data set only visit 5 is a flat, so specifying "--id visit=5" is all you need to specify for the butler
+  to find our flat. In case you  would want to reduce all Flats taken 2015-12-22 for spectrograph 2, red arm,
```
jdswinbank: (Don't need the "would" — just "in case you want to" is fine)

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+
+constructDark.py $PFS_DATA --rerun USERNAME/tmp --id field=DARK dateObs=2015-12-22 arm=r spectrograph=2 --calibId calibVersion=dark calibDate=2015-12-22 arm=r spectrograph=2 --cores 1
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Having the master Bias and Dark, we can now create our master Flat. Currently the master Flat is only used to 
+  trace the apertures of the fiber traces and to calculate the spatial profile for the optimal extraction.
+  Note that for the actual flat-fielding dithered Flats will be used in the near future, and this kind of Flat
+  here will be renamed to apDef (aperture definition).
+  In our data set only visit 5 is a flat, so specifying "--id visit=5" is all you need to specify for the butler
```
jdswinbank: Again with the butler!

doc/install.text
```diff
@@ -0,0 +1,197 @@
...
+  images are valid for 180 days. We will need to repeat this step every time we create a new calibration
+  image for the butler to find it when its use is required by a later called task.
+
+genCalibRegistry.py --root $PFS_DATA/CALIB --camera PFS --validity 180
+
+
+* Now we can create a trimmed and scaled, Bias-subtracted master Dark and ingest that into our calibration registry.
+  Again, you need to substitute USERNAME with your name.
+  If you don't want Dark subtraction you can skip this step, but remember further down to turn Dark subtraction
+  off.
```
jdswinbank: I'd drop that last sentence: it's unnecessary complexity.

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -264,7 +261,7 @@ def run(self, expRefList, butler, wLenFile=None, lineList=None, immediate=True):
             else:
                 print 'setSpectrum for spectrumSetFromProfile[',i,'] failed'
 
-        if True:
+        if False:
```
jdswinbank: Hmm. LSST's [RFC-205](https://jira.lsstcorp.org/browse/RFC-205) for a discussion of this. It looks like they're converging on a decision to discourage this idiom, and we've agreed to follow LSST standards... but they haven't quite converged yet, so maybe you can get away with it for now. However, I definitely think we'd require a comment explaining what's going on here.

tests/FiberTraces.py
```diff
@@ -30,10 +30,10 @@ class FiberTraceTestCase(tests.TestCase):
 
     def setUp(self):
         drpStellaDataDir = lsst.utils.getPackageDir("drp_stella_data")
-        flatfile = os.path.join(drpStellaDataDir,"data/PFS/CALIB/FLAT/r/flat/pfsFlat-2016-01-12-0-2r.fits")
+        flatfile = os.path.join(drpStellaDataDir,"tests/data/PFS/CALIB/FLAT/r/flat/pfsFlat-2016-01-12-0-2r.fits")
```
jdswinbank: Should really have a space after the `,` here (and on all the lines below).

doc/install.text
ziggyman: Mac is fine

doc/install.text
ziggyman: Done

doc/install.text
ziggyman: Fixed

doc/install.text
ziggyman: Removed everything connected to sconsUtils

doc/install.text
ziggyman: Removed

doc/install.text
ziggyman: Added a section at the end how to setup the pipeline again later

doc/install.text
ziggyman: Removed the --show config=glob for now as I don't really understand it yet myself

doc/install.text
ziggyman: fixed

doc/install.text
ziggyman: Correct, this one is for the data and the previous one was where we built the code.

doc/install.text
ziggyman: Adopted

doc/install.text
ziggyman: Changed the sentence and removed "-L warn" parameter but left it in the description as an option for the user

doc/install.text
ziggyman: Added link to pipe_base's Doxygen

doc/install.text
ziggyman: At least on my Mac it sometimes fails with a segfault if run in parallel. I'd keep this in for now until we sorted this out...

doc/install.text
ziggyman: filed ticket DM-7522

doc/install.text
ziggyman: Reworded

doc/install.text
ziggyman: Sentence dropped

doc/install.text
ziggyman: Reworded

doc/install.text
ziggyman: "would" removed

doc/install.text
ziggyman: Sentence dropped as no Dark subtraction option removed from the quick-start guide

doc/install.text
ziggyman: done

doc/install.text
ziggyman: Reworded

doc/install.text
ziggyman: dropped "would"

doc/install.text
ziggyman: Added link to the data model which describes the output file

python/pfs/drp/stella/reduceArcTask.py
ziggyman: Added comment

tests/FiberTraces.py
ziggyman: done

PR #4, opened 2016-11-04
Review for PIPE2D-109

jdswinbank:


jdswinbank:
Please try to keep the first line of commit messages to <= 50 chars, so that they're easier to see & read, both in the display above & on the command line.

src/math/CurveFitting.cc
```diff
@@ -283,7 +283,7 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
           D_A1_Y.push_back(D_A1_Y_In[i_pos]);
           if (B_HaveMeasureErrors)
             D_A1_MeasureErrors.push_back(D_A1_MeasureErrorsBak[i_pos]);
-          I_A1_OrigPos.push_back(D_A1_Y_In[i_pos]);
+          I_A1_OrigPos.push_back(i_pos);
```
jdswinbank: Coming back to this later: I see there is now a test for ``PolyFit``: great!

tests/Spectra.py
```diff
@@ -48,7 +48,7 @@ def setUp(self):
         
         self.nFiberTraces = 11
         self.nRowsPrescan = 49
-        self.maxRMS = 0.053
+        self.maxRMS = 0.067
```
jdswinbank: If I've understood the commit message, then this number seems to be pretty poorly justified: it's just a value you happened to measure one day. I guess that's better than nothing (at the least, it serves as a regression test to demonstrate that changes to the code elsewhere haven't made it worse), but you should add a comment to that effect.

tests/Spectra.py
```diff
@@ -454,8 +454,7 @@ def testWavelengthCalibrationWithRefSpec(self):
                 self.assertLess(spec.getDispRms(), self.maxRMS)
             
     def testWavelengthCalibrationWithoutRefSpec(self):
-        if True:
-            print "testing wavelength calibration"
+        if False:
```
jdswinbank: My understanding is that you're planning to re-activate these lines when the wavelength-per-pixel file is updated. Is that correct? If so:

- Why did you merge the new arc without the new wavelength per pixel? It seems like this should have been atomic.
- Please file an issue, triggered by the issue requesting the new wavelength-per-pixel file, to re-enable these tests;
- Please add a comment explaining what's been done.

Also, why is this in the same commit as the change above? It's a different piece of work.

include/pfs/drp/stella/SpectraTemplates.hpp
```diff
@@ -532,10 +529,16 @@ bool pfs::drp::stella::Spectrum<SpectrumT, MaskT, VarianceT, WavelengthT>::ident
...
-    #endif
+    _mask.addMaskPlane("REJECTED_LINES");
+    unsigned short maskVal = 1 << _mask.getMaskPlane("REJECTED_LINES");
+    for (int i = 0; i < rejected->size(); ++i){
+        #ifdef __DEBUG_IDENTIFY__
+            cout << "identify: rejected D_A1_FittedPos[" << (*rejected)[i] << "] = " << D_A1_FittedPos[(*rejected)[i]] << endl;
+        #endif
+        for (int p = (D_A1_FittedPos[(*rejected)[i]-2] < 0 ? 0 : D_A1_FittedPos[(*rejected)[i]-2]);
+                 p < (D_A1_FittedPos[(*rejected)[i]+2] >= _length ? _length-1 : D_A1_FittedPos[(*rejected)[i]+2]); ++p)
+            _mask(p, 0) |= maskVal;
```
jdswinbank: The commit messages says you're just changing "_mask in Spectrum from ndarray to afwImage::Mask", but this looks very much like you're actually setting bits in the mask.  This should probably be a separate commit; if it isn't, then at the least this change should be mentioned in the commit message.

include/pfs/drp/stella/SpectraTemplates.hpp
```diff
@@ -503,37 +503,86 @@ bool pfs::drp::stella::Spectrum<SpectrumT, MaskT, VarianceT, WavelengthT>::ident
 
     ndarray::Array< double, 1, 1 > D_A1_FittedWLenCheck = math::getSubArray( D_A1_WLen, I_A1_IndexCheckPos );
 
+    std::vector<string> S_A1_Args(3);
```
jdswinbank: I'm confused as to why all this code is in a header file: is that really necessary?

jdswinbank: How do you know this code works? I don't see any changes to the test suite beyond those testing the narrow functionality of the polyfit routine itself. Can you add a test which demonstrates that it is being applied here properly?

jdswinbank:
Hmm. GitHub appears to have lost some of my comments in that review. Give me a few minutes to try and figure out what happened...

tests/Math.py
```diff
@@ -33,7 +33,26 @@ def testWhere(self):
...
+        print 'type(xArr[0]) = ',type(xArr[0])
+        nDeg = 2#c_size_t(3)
+        print 'type(nDeg) = ',type(nDeg)
+        coeffs = drpStella.PolyFit(xArr, yArr, nDeg)
+        print 'coeffs = ',coeffs
+        
+        coeffs = drpStella.PolyFit(xArr, yArr, nDeg, -3., 3., 2)
+        print 'coeffs = ',coeffs
+
+        self.assertTrue(drpStella.testPolyFit())
```
jdswinbank: I'm confused by what's going on here.

It seems like you first do some calculations in Python, but never check their results. Then you delegate the whole of the test to C++ code. Is that correct? If so, please don't.

First of all, drop the printing unless it's useful.

Next, please check the coeffs returned in Python for validity.

At least some of the tests in drpStella.testPolyFit seem to be degenerate with the Python code (for example, the Python call at line 48 above seems to be basically equivalent to the C++ at line 425 of Utils.cc). Please drop all of these redundant tests.

If it's possible to test all of the PolyFit functionality in pure Python code, please do that. If it's not possible, you can add a minimal C++ test which just exercises the functionality which isn't exposed to Python, and add comments explaining 1) that that's what you're doing, 2) why the relevant functionality isn't available in Python.

src/math/CurveFitting.cc
```diff
@@ -235,64 +212,89 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
+      #endif
       for (size_t i_pos=0; i_pos < D_A1_Y_In.getShape()[0]; i_pos++){
         D_Dev = D_A1_Y_In[i_pos] - D_A1_PolyRes[i_pos];
+        #ifdef __DEBUG_POLYFIT__
+          cout << "pfs::drp::stella::math::CurveFitting::PolyFit: i_pos = " << i_pos << ": D_Dev = " << D_Dev << ", D_SDev = " << D_SDev << endl;
+        #endif
         if (((D_Dev < 0) && (D_Dev > (D_LReject_In * D_SDev))) || 
             ((D_Dev >= 0) && (D_Dev < (D_UReject_In * D_SDev)))){
-          D_A1_X.push_back(xNew[i_pos]);
+          D_A1_X.push_back(D_A1_X_In[i_pos]);
```
jdswinbank: The fix is simple and unarguable. The code is long and hard to read, so I've not fought through all the implications, but this certainly looks right.

But: this seems like something which should be easily caught in a unit test. I wonder why we weren't hitting it before? Can you add a test to demonstrate that this code now does the right thing?

Updated later to add: I see there is now a test for PolyFit: great! Can you confirm that it would have caught this problem?

src/Spectra.cc
```diff
@@ -423,7 +423,7 @@ ndarray::Array< int, 2, 1 > pfsDRPStella::SpectrumSet<SpectrumT, MaskT, Variance
       );
     }
 
-    mask[ ndarray::view( yLow, yHigh + 1 )( iFiber ) ] = spectrum->getMask()[ ndarray::view() ];
+//    mask[ ndarray::view( yLow, yHigh + 1 )( iFiber ) ] = spectrum->getMask().getArray()[ ndarray::view() ];
```
jdswinbank: Why are you leaving this commented out?

jdswinbank:
There seems to be various changes in 1777877264fe803197841dd49302edea5f28d7fe that go beyond the statement in the commit message "Turns out it was just a ' instead of " in cout statement in Utils.cc": some of them seem to be logic changes to the code. Please expand the commit message to explain what's going on (or move these changes to a separate commit).

jdswinbank:
I think I've now retrieved all my comments. Sorry for the confusion — clearly, I need to learn how to operate GitHub — and I hope it's now clear. If it's not, please let me know.

include/pfs/drp/stella/SpectraTemplates.hpp
```diff
@@ -532,10 +529,16 @@ bool pfs::drp::stella::Spectrum<SpectrumT, MaskT, VarianceT, WavelengthT>::ident
     #endif
     
     /// Remove lines rejected by PolyFit from D_A1_FittedPos and D_A1_FittedWLen
-    #ifdef __DEBUG_IDENTIFY__
-        for (int i = 0; i < rejected->size(); ++i)
-            cout << "identify: rejected[" << i << "] = " << (*rejected)[i] << endl;
-    #endif
+    _mask.addMaskPlane("REJECTED_LINES");
```
ziggyman: Where should that go then? Into the constructor of Spectrum? Or should I add another parameter to Spectrum::identify?

ziggyman: Moved to Spectrum standard constructor

include/pfs/drp/stella/SpectraTemplates.hpp
ziggyman: split into 2 commits

tests/Spectra.py
ziggyman: Added comment

tests/Spectra.py
ziggyman: re-enabled test

tests/Math.py
ziggyman: All tests are now in C++. Added comment why

src/math/CurveFitting.cc
```diff
@@ -208,25 +194,16 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
-    std::vector<T> D_A1_X(xNew.begin(), xNew.end());
+    std::vector<T> D_A1_X(D_A1_X_In.begin(), D_A1_X_In.end());
     std::vector<T> D_A1_Y(D_A1_Y_In.begin(), D_A1_Y_In.end());
-    std::vector<T> D_A1_X_New(D_A1_X.size());
-    std::vector<T> D_A1_Y_New(D_A1_Y.size());
-    std::vector<T> D_A1_MeasureErrors(D_A1_X.size());
-    std::vector<T> D_A1_MeasureErrors_New(D_A1_X.size());
+//    std::vector<T> D_A1_X_New(D_A1_X.size());
+//    std::vector<T> D_A1_Y_New(D_A1_Y.size());
+//    std::vector<T> D_A1_MeasureErrors_New(D_A1_X.size());
```
ziggyman: removed

include/pfs/drp/stella/utils/Utils.h
```diff
@@ -184,5 +186,7 @@ namespace pfs { namespace drp { namespace stella { namespace utils{
   
   template< typename T >
   PTR( T ) getPointer( T & );
+  
+  bool testPolyFit();
```
ziggyman: Added comment

tests/Math.py
```diff
@@ -33,7 +33,26 @@ def testWhere(self):
         for i in range(10):
             if i != 5:
                 self.assertEqual(int1DArrRes[i], 0)
-            
+                
+    def testPolyFit(self):
```
ziggyman: moved setUp into testWhere

tests/Math.py
```diff
@@ -48,7 +48,7 @@ def testPolyFit(self):
         coeffs = drpStella.PolyFit(xArr, yArr, nDeg)
         print 'coeffs = ',coeffs
         
-        coeffs = drpStella.PolyFit(xArr, yArr, nDeg, -2., 2., 2)
+        coeffs = drpStella.PolyFit(xArr, yArr, nDeg, -3., 3., 2)
```
ziggyman: Removed python tests, all tests are now done in C++

src/math/CurveFitting.cc
```diff
@@ -435,13 +445,18 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
+      D_A1_XRange[0] = -1.;
+      D_A1_XRange[1] = 1.;
+      xNew = D_A1_X_In;
     }
-    xNew = pfs::drp::stella::math::convertRangeToUnity(D_A1_X_In,
-                                                       D_A1_XRange);
+    #ifdef __DEBUG_POLYFIT__
+      cout << "PolyFit: xNew = " << xNew << endl;
+      cout << "++++++++++++++++++++++++++++++++++++" << endl;
+    #endif
```
ziggyman: Added more tests for PolyFit in C++

src/math/CurveFitting.cc
```diff
@@ -208,18 +198,20 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
         cout << "pfs::drp::stella::math::CurveFitting::PolyFit: *P_D_A1_XRange set to " << *P_D_A1_XRange << endl;
       #endif
     }
-    ndarray::Array<T, 1, 1> xNew = pfs::drp::stella::math::convertRangeToUnity(D_A1_X_In, xRange);
 //    #ifdef __DEBUG_POLY__
       cout << "pfs::drp::stella::math::CurveFitting::Poly: D_A1_X_In = " << D_A1_X_In << endl;
       cout << "pfs::drp::stella::math::CurveFitting::Poly: xNew = " << xNew << endl;
 //    #endif
+    #ifdef __DEBUG_POLYFIT__
+      cout << "pfs::drp::stella::math::CurveFitting::PolyFit: xRange = " << xRange << endl;
```
ziggyman: Replaced all __DEBUG__ messages in CurveFitting.cc with logger messages

src/math/CurveFitting.cc
```diff
@@ -156,10 +156,6 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
       P_I_A1_Rejected.reset();
       P_I_A1_Rejected = *((PTR(std::vector<size_t>)*)(ArgV[I_Pos]));
       #ifdef __DEBUG_POLYFIT__
-        cout << "PolyFit: P_I_A1_Rejected = ";
-        for (auto it = P_I_A1_Rejected->begin(); it != P_I_A1_Rejected->end(); ++it)
-          cout << *it << " ";
-        cout << endl;
```
ziggyman: split up commit and changed all debugging messages to logger messages

include/pfs/drp/stella/Controls.h
```diff
@@ -487,7 +487,9 @@ struct DispCorControl {
     LSST_CONTROL_FIELD( stretchMinLength, int, "Minimum length to stretched pieces to (< lengthPieces)" );
     LSST_CONTROL_FIELD( stretchMaxLength, int, "Maximum length to stretched pieces to (> lengthPieces)" );
     LSST_CONTROL_FIELD( nStretches, int, "Number of stretches between <stretchMinLength> and <stretchMaxLength>");
-    LSST_CONTROL_FIELD( verticalPrescanHeight, int, "Number of rows in the raw image containing the vertical prescan");
+    LSST_CONTROL_FIELD( verticalPrescanHeight, int, "Number of rows in the raw image pixel vs. wavelength containing the vertical prescan");
```
ziggyman: fixed

src/math/CurveFitting.cc
```diff
@@ -113,8 +113,6 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
       throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
     }
 
-    int I_NReject = 0;
-    int I_DataValues_New = 0;
```
ziggyman: done

src/math/CurveFitting.cc
```diff
@@ -89,13 +84,13 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
                                            I_Degree_In,
-                                           D_Reject_In,
+                                           0.-D_Reject_In,
                                            D_Reject_In,
                                            -1,
                                            S_A1_Args_In,
                                            ArgV);
   }
-
+
```
ziggyman: moved into commit only consisting of removal of trailing spaces

jdswinbank:
Please squash 49629011b5424232db7d527cf60af7ab1bbc0534 into 43f02280ad2393ab7746ba3a334f7e7bd0d8d424

jdswinbank:
Please expand the commit message for 43f02280ad2393ab7746ba3a334f7e7bd0d8d424 to indicate why you made this change.

src/math/CurveFitting.cc
```diff
@@ -123,75 +119,53 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
-      #ifdef __DEBUG_POLYFIT__
-        cout << "PolyFit: P_I_A1_NotRejected = ";
-        for (auto it = P_I_A1_Rejected->begin(); it != P_I_A1_Rejected->end(); ++it)
-          cout << *it << " ";
-        cout << endl;
-        cout << "pfs::drp::stella::math::CurveFitting::PolyFit: KeyWord REJECTED read" << endl;
-      #endif
+      LOGLS_DEBUG(_log, "KeyWord REJECTED read");
     }
+    P_I_A1_Rejected->resize(0);
```
jdswinbank: Why has it only become necessary to initialize this (and ``P_I_A1_NotRejected``, ``P_I_NRejected``) now? Is it purely cosmetic, or is this fixing a bug?

If it's fixing a bug, please add a test case.

src/math/CurveFitting.cc
```diff
@@ -200,41 +174,64 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
         throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
       }
       xRange.deep() = *P_D_A1_XRange;
-      #ifdef __DEBUG_POLYFIT__
-        cout << "pfs::drp::stella::math::CurveFitting::PolyFit: *P_D_A1_XRange set to " << *P_D_A1_XRange << endl;
-      #endif
+      LOGLS_DEBUG(_log, "*P_D_A1_XRange set to " << *P_D_A1_XRange);
     }
-    ndarray::Array<T, 1, 1> xNew;
+    ndarray::Array<T, 1, 1> xNew = ndarray::allocate(D_A1_X_In.getShape()[0]);
```
jdswinbank: This is explicitly flagged as a bug fix. Where is the unit test?

src/math/CurveFitting.cc
```diff
@@ -477,13 +472,11 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
         message += to_string(P_D_A1_Sigma->getShape()[0]) +") != nCoeffs(=" + to_string(nCoeffs) + ")";
-        cout << message << endl;
         throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
       }
-      #ifdef __DEBUG_POLYFIT__
-        cout << "pfs::drp::stella::math::CurveFitting::PolyFit: KeyWord_Set(SIGMA): *P_D_A1_Sigma set to " << (*P_D_A1_Sigma) << endl;
-      #endif
+      LOGLS_DEBUG(_log, "KeyWord_Set(SIGMA): *P_D_A1_Sigma set to " << (*P_D_A1_Sigma));
     }
+    P_D_A1_Sigma->deep() = 0.;
```
jdswinbank: Again, why is it only necessary to initialize this now? Please explain the motivation. If it's a bug fix, can we have a unit test?

src/math/CurveFitting.cc
```diff
@@ -495,215 +488,115 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
         message += to_string(P_D_A2_Covar->getShape()[1]) + ") != nCoeffs(=" + to_string(nCoeffs) + ")";
-        cout << message << endl;
         throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
       }
-      #ifdef __DEBUG_POLYFIT__
-        cout << "pfs::drp::stella::math::CurveFitting::PolyFit: KeyWord_Set(COVAR): *P_D_A2_Covar set to " << (*P_D_A2_Covar) << endl;
-      #endif
+      LOGLS_DEBUG(_log, "KeyWord_Set(COVAR): *P_D_A2_Covar set to " << (*P_D_A2_Covar));
     }
+    P_D_A2_Covar->deep() = 0.;
```
jdswinbank: Again, why is it only necessary to initialize this now? Please explain the motivation. If it's a bug fix, can we have a unit test?

src/math/CurveFitting.cc
```diff
@@ -243,59 +240,80 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
         if (((D_Dev < 0) && (D_Dev > (D_LReject_In * D_SDev))) || 
             ((D_Dev >= 0) && (D_Dev < (D_UReject_In * D_SDev)))){
-          D_A1_X.push_back(xNew[i_pos]);
+          D_A1_X.push_back(D_A1_X_In[i_pos]);
           D_A1_Y.push_back(D_A1_Y_In[i_pos]);
           if (B_HaveMeasureErrors)
-            D_A1_MeasureErrors.push_back(D_A1_MeasureErrorsBak[i_pos]);
-          I_A1_OrigPos.push_back(D_A1_Y_In[i_pos]);
+            V_MeasureErrors.push_back((*P_D_A1_MeasureErrors)[i_pos]);
+          I_A1_OrigPos.push_back(i_pos);
```
jdswinbank: I note for posterity that this is the sort of thing for which I'd expect the fix to be accompanied by a unit test. However, since we effectively don't have tests at this point (they are added by a future commit: 👍), we can let this slide.

src/math/CurveFitting.cc
```diff
@@ -200,41 +174,64 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
-    std::vector<T> D_A1_Y_New(D_A1_Y.size());
-    std::vector<T> D_A1_MeasureErrors(D_A1_X.size());
-    std::vector<T> D_A1_MeasureErrors_New(D_A1_X.size());
     std::vector<size_t> I_A1_OrigPos(D_A1_X_In.getShape()[0]);
+    std::vector<T> V_MeasureErrors(P_D_A1_MeasureErrors->begin(), P_D_A1_MeasureErrors->end());
+    LOGLS_DEBUG(_log, "V_MeasureErrors = " << V_MeasureErrors);
+
+    PTR(ndarray::Array<T, 1, 1>) P_D_A1_MeasureErrorsTemp(new ndarray::Array<T, 1, 1>(ndarray::external(V_MeasureErrors.data(), 
+                                                                                                        ndarray::makeVector(int(V_MeasureErrors.size())), 
+                                                                                                        ndarray::makeVector(1))));
```
jdswinbank: Remember the 110 character line length limit. Also, please don't leave extra white space at the end of the line (the above end with comma-space-newline).

jdswinbank: For all the changes on this commit (077f70ad770d169bc107de72387b5d14675ab678), more comments would be helpful. The commit message provides a useful roadmap to what's happening (thank you!), but a few signposts to make the code easier to follow would be even better.

PR #16, opened 2017-01-18
Tickets/pipe2 d 77

ziggyman:


tests/Spectra.py
```diff
@@ -371,6 +369,49 @@ def testGetSpectra(self):
...
+            ratio = orig / np.where(rec < 0.1, 1., rec)
+
+            binWidth = 1. / (myProfileTask.getFiberTraceProfileFittingControl().overSample
+                             * nBinsPerOverSampleStep)
+            distanceMeanStd = drpStellaMath.getMeanStdXBins(dist,
+                                                            ratio,
+                                                            binWidth)
+            print 'type(distanceMeanStd) = ',type(distanceMeanStd)
+            for iBin in range(distanceMeanStd.shape[0]):
+                self.assertTrue((distanceMeanStd[iBin, 1, 0] - distanceMeanStd[iBin, 1, 1] < 1.)
```
PaulPrice: Maybe it's because the meanings are hidden behind indices in the `distanceMeanStd` (see previous comment on `getMeanStdXBins`), but this isn't clear to me.  The docstring says "The test passes if 1.0 is within 1 standard deviation of the mean value", so I would have expected something like `self.assertTrue(numpy.all(numpy.fabs(mean - 1.0) < std))`.
And it's not clear to me why the test criterion you've chosen is a good one.  Why don't you want to just check that the standard deviation is small and the mean is even smaller?

src/FiberTraces.cc
```diff
@@ -1047,6 +1047,8 @@ namespace pfsDRPStella = pfs::drp::stella;
       throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
     }
 
+    double tmpVal;
```
PaulPrice: You shouldn't declare a variable until you're going to use it, and you should initialise it at the same time as you declare it.

src/FiberTraces.cc
```diff
@@ -1253,14 +1259,20 @@ namespace pfsDRPStella = pfs::drp::stella;
...
-        if (profArraySwath[iRow][iCol] < 0.)
-          profArraySwath[iRow][iCol] = 0.;
+        /// The spline's knots are calculated from bins in x centered at the
+        /// oversampled positions in x.
+        /// Outside the range in x on which the spline is defined, which is
+        /// [min(xRange) + overSample/2., max(xRange) - overSample/2.], so to
+        /// say in the outer (half overSample), the spline is extrapolated from
+        /// the 1st derivative at the end points.
+        tmpVal = spline(xArray[iRow][iCol]);
+        profArraySwath[iRow][iCol] = (tmpVal >= 0. ? tmpVal : 0.);
```
PaulPrice: Is the above comment supposed to be explaining the need to clip negative values?
Are there other kinds of splines that would behave better at the ends?

tests/FiberTraces.py
```diff
@@ -687,7 +687,22 @@ def testFiberTraceSetFunctions(self):
...
+    def testAddFiberTraceToCcdImage(self):
+        fts = drpStella.findAndTraceAperturesF(self.flat.getMaskedImage(), self.ftffc)
+        origMaskedImage = self.flat.getMaskedImage()
+        imageShape = origMaskedImage.getImage().getArray().shape
+        sumImage = afwImage.ImageF(np.zeros(shape=imageShape, dtype=np.float32))
+        ftsOrig = fts
+        for iFt in range(fts.size()):
+            ft = fts.getFiberTrace(iFt)
+            drpStella.addFiberTraceToCcdImage(ft, ft.getImage(), sumImage)
+        self.assertTrue(fts.createTraces(afwImage.makeMaskedImage(sumImage)))
```
PaulPrice: Does this modify `fts`?  Otherwise, I don't see the point in the next loop.  But if it does, that's kinda surprising because it's already a `FiberTraceSet` and I wouldn't expect it to change.

include/pfs/drp/stella/math/Math.h
```diff
@@ -583,6 +583,13 @@ namespace pfs { namespace drp { namespace stella {
                                      ndarray::Array< U, 2, 1 > const& valuesIfTrue,
                                      ndarray::Array< U, 2, 1 > const& valuesIfFalse );
     
+    template< typename T, typename U >
+    ndarray::Array< U, 2, 1 > where( ndarray::Array< T, 2, 1 > const& arrayToCompareTo,
+                                     std::string const& op,
+                                     T const valueToCompareTo, 
+                                     ndarray::Array< U, 2, 1 > const& valuesIfTrue,
+                                     ndarray::Array< U, 2, 1 > const& valuesIfFalse );
+
```
ziggyman: Good point that it matches the declaration before. I guess something went wrong with a rebase. I removed this commit.
I filed a ticket [PIPE2D-156](https://pfspipe.ipmu.jp/jira/browse/PIPE2D-156) to remove all `where` function declarations from `stellaLib.i` as `numpy.where` should be used instead.
Regarding the speed in C++, in the majority of use cases the arrays `valuesIfTrue` and `valuesIfFalse` exist already, so they won't have to be created just to call the `where` function

include/pfs/drp/stella/FiberTraces.h
```diff
@@ -420,7 +420,21 @@ namespace math{
...
+     *                 each row from <smallArr> in
+     * @param yMin in: row number in which to start adding <smallArr> into
+     * @param bigArr in/out: Array to which to add <smallArr>
+     */
+    template< typename smallT, typename bigT, int I, int J >
+    void addArrayIntoArray( ndarray::Array< smallT, 2, I > const& smallArr,
+                            ndarray::Array< size_t, 2, 1 > const& xMinMax,
+                            size_t const& yMin,
+                            ndarray::Array< bigT, 2, J > & bigArr );
+}
```
ziggyman: The function is there to insert a curved `FiberTrace` into a CCD array. I added a comment clarifying this.

src/FiberTraces.cc
ziggyman: Moved the declaration down to just before the `for` loops where it will be used, and initialized it. I don't want to move the declaration inside the `for` loops as it would cause multiple memory allocations and deallocations.

src/FiberTraces.cc
```diff
@@ -668,32 +668,36 @@ namespace pfsDRPStella = pfs::drp::stella;
...
   /// Return shared pointer to an image containing the reconstructed 2D spectrum of the FiberTrace
   template<typename ImageT, typename MaskT, typename VarianceT>
-  afwImage::Image<double> pfsDRPStella::FiberTrace<ImageT, MaskT, VarianceT>::getReconstructed2DSpectrum(const pfsDRPStella::Spectrum<ImageT, MaskT, VarianceT, VarianceT> & spectrum) const{
-    ndarray::Array<double, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(), _trace->getWidth());
+  PTR( afwImage::Image< ImageT > ) pfsDRPStella::FiberTrace<ImageT, MaskT, VarianceT>::getReconstructed2DSpectrum(const pfsDRPStella::Spectrum<ImageT, MaskT, VarianceT, VarianceT> & spectrum) const{
+    ndarray::Array<ImageT, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(), _trace->getWidth());
     int i_row = 0;
-    for (auto itProf = _profile->getArray().begin(), itRec = F_A2_Rec.begin(); itProf != _profile->getArray().end(); ++itProf, ++itRec, ++i_row)
+    auto itRec = F_A2_Rec.begin();
+    for (auto itProf = _profile->getArray().begin(); itProf != _profile->getArray().end(); ++itProf, ++itRec, ++i_row)
```
ziggyman: if I do this I get the error:
```
/Users/azuri/anaconda/envs/lsst-v12_1/opt/lsst/ndarray/include/ndarray/ExpressionTraits.h:84:5: error: static_assert failed "(ND::value == ExpressionTraits<Operand2>::ND::value)"
    BOOST_STATIC_ASSERT((ND::value == ExpressionTraits<Operand2>::ND::value));
    ^                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/azuri/anaconda/envs/lsst-v12_1/opt/lsst/boost/include/boost/static_assert.hpp:70:41: note: expanded from macro 'BOOST_STATIC_ASSERT'
#     define BOOST_STATIC_ASSERT( ... ) static_assert(__VA_ARGS__, #__VA_ARGS__)
                                        ^
/Users/azuri/anaconda/envs/lsst-v12_1/opt/lsst/ndarray/include/ndarray/ExpressionBase.h:43:22: note: in instantiation of template class 'ndarray::ExpressionTraits<ndarray::detail::BinaryOpExpression<ndarray::Array<double, 2, 1>, ndarray::Array<float, 1, 1>, ndarray::detail::MultipliesTag::ScalarFunction<double, float>, 2> >' requested here
    typedef typename ExpressionTraits<Derived>::Element Element;
                     ^
/Users/azuri/anaconda/envs/lsst-v12_1/opt/lsst/ndarray/include/ndarray/detail/BinaryOp.h:91:35: note: in instantiation of template class 'ndarray::ExpressionBase<ndarray::detail::BinaryOpExpression<ndarray::Array<double, 2, 1>, ndarray::Array<float, 1, 1>, ndarray::detail::MultipliesTag::ScalarFunction<double, float>, 2> >' requested here
class BinaryOpExpression : public ExpressionBase< BinaryOpExpression<Operand1,Operand2,BinaryFunction,N> > {
                                  ^
src/FiberTraces.cc:672:55: note: in instantiation of template class 'ndarray::detail::BinaryOpExpression<ndarray::Array<double, 2, 1>, ndarray::Array<float, 1, 1>, ndarray::detail::MultipliesTag::ScalarFunction<double, float>, 2>' requested here
    auto F_A2_Rec = ndarray::copy(_profile->getArray()*spectrum.getSpectrum());
                                                      ^
src/FiberTraces.cc:2869:30: note: in instantiation of member function 'pfs::drp::stella::FiberTrace<float, unsigned short, float>::getReconstructed2DSpectrum' requested here
template class pfsDRPStella::FiberTrace<float, unsigned short, float>;
```

PaulPrice: Maybe that's because it's multiplying a `float` array and a `double` array, or because the dimensions are different.  I think it's worth exploring, not just to save the `for` loop here, but so we'll know how to do it in other cases.

PaulPrice:
@ziggyman, I see a bunch of new commits. Should I review everything, or just starting from 9301375?

ziggyman:
starting with 9301375 is definitely okay.

src/FiberTraces.cc
PaulPrice: I think you're probably getting that anyway.

src/FiberTraces.cc
ziggyman: Right, `spectrum` could be a `float` or `double` array, and `profile` is a `double` `ndarray`. `Eigen` would require an explicit cast for this, while` ndarray` does not. After consultation with Jim we agreed that a `for` loop for each row and then multiplying each row with a scalar is probably the best solution:
```
    ndarray::Array<ImageT, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(), _trace->getWidth());
    auto itRec = F_A2_Rec.begin();
    auto itSpec = spectrum.getSpectrum().begin();
    for (auto itProf = _profile->getArray().begin(); itProf != _profile->getArray().end(); ++itProf, ++itRec, ++itSpec)
      (*itRec) = (*itProf) * (*itSpec);
```
For the next function `getReconstructedBackground` please refer to the next comment...

src/FiberTraces.cc
```diff
@@ -668,32 +668,36 @@ namespace pfsDRPStella = pfs::drp::stella;
...
+    PTR( afwImage::Image< ImageT > ) imagePtr( new afwImage::Image< ImageT >( F_A2_Rec ) );
+    return imagePtr;
   }
 
   /// Return shared pointer to an image containing the reconstructed background of the FiberTrace
   template<typename ImageT, typename MaskT, typename VarianceT>
-  afwImage::Image<double> pfsDRPStella::FiberTrace<ImageT, MaskT, VarianceT>::getReconstructedBackground( const pfsDRPStella::Spectrum<ImageT, MaskT, VarianceT, VarianceT> & background) const{
-    ndarray::Array<double, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(), _trace->getWidth());
+  PTR(afwImage::Image<ImageT>) pfsDRPStella::FiberTrace<ImageT, MaskT, VarianceT>::getReconstructedBackground( const pfsDRPStella::Spectrum<ImageT, MaskT, VarianceT, VarianceT> & background) const{
+    ndarray::Array<ImageT, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(), _trace->getWidth());
```
ziggyman: Here we multiply each row with a scalar, or a matrix with a vector, just like the function before (`getReconstructed2DSpectrum`). The difference is that here it is guaranteed that both matrix and vector have the same type (float or double). In this case it is easy and efficient to use the cross-product in `eigen`:
```
    ndarray::Array<ImageT, 1, 1> oneRow = ndarray::allocate(_trace->getWidth());
    oneRow.deep() = 1.;
    ndarray::Array<ImageT, 2, 1> F_A2_Rec = ndarray::allocate(_trace->getHeight(),
                                                              _trace->getWidth());
    F_A2_Rec.asEigen() = background.getSky().asEigen() * oneRow.asEigen().transpose();
```

src/FiberTraces.cc
```diff
@@ -2716,16 +2716,18 @@ namespace pfsDRPStella = pfs::drp::stella;
...
-      }
-      return true;
+      ndarray::Array<size_t, 2, 1> xMinMax = ndarray::allocate(minCenMax.getShape()[0], 2);
+      xMinMax[ndarray::view()(0)] = minCenMax[ndarray::view()(0)];
+      xMinMax[ndarray::view()(1)] = minCenMax[ndarray::view()(2)];
+      addArrayIntoArray(fiberTraceRepresentation.getArray(),
+                        xMinMax,
+                        y,
+                        ccdArray);
+      return;
```
ziggyman: Point taken, but I leave it as it is for now. If you think it should get changed I can file a ticket...?

PaulPrice: Nah.  You've done the work, so let's leave it for now.

src/FiberTraces.cc
```diff
@@ -2898,67 +2900,67 @@ template PTR(pfsDRPStella::FiberTrace< double, unsigned short, float >) pfsDRPSt
                                                                                                  double,
                                                                                                  size_t );
 
-template bool pfsDRPStella::math::addFiberTraceToCcdArray( pfsDRPStella::FiberTrace< float, unsigned short, float > const&,
+template void pfsDRPStella::math::addFiberTraceToCcdArray( pfsDRPStella::FiberTrace< float, unsigned short, float > const&,
```
ziggyman: Hmmm... Could you give me an example please? Couldn't find one on the internet doing a quick search...

src/FiberTraces.cc
ziggyman: Negative profile values are clipped because they are not physical.
I added another comment clarifying this.

src/FiberTraces.cc
ziggyman: Ah okay, found one below... Great, thanks! :)

src/FiberTraces.cc
```diff
@@ -2946,3 +2961,36 @@ template bool pfsDRPStella::math::addFiberTraceToCcdImage( pfsDRPStella::FiberTr
...
+                                                     size_t const&,
+                                                     ndarray::Array< float, 2, 2 > & );
+template void pfsDRPStella::math::addArrayIntoArray( ndarray::Array< float const, 2, 2 > const&,
+                                                     ndarray::Array< size_t, 2, 1 > const&,
+                                                     size_t const&,
+                                                     ndarray::Array< float, 2, 1 > & );
+template void pfsDRPStella::math::addArrayIntoArray( ndarray::Array< float const, 2, 2 > const&,
+                                                     ndarray::Array< size_t, 2, 1 > const&,
+                                                     size_t const&,
+                                                     ndarray::Array< float, 2, 2 > & );
```
ziggyman: Great, done!

python/pfs/drp/stella/createFlatFiberTraceProfileTask.py
```diff
@@ -51,20 +51,24 @@ class CreateFlatFiberTraceProfileTask(Task):
 
     def __init__(self, *args, **kwargs):
         super(CreateFlatFiberTraceProfileTask, self).__init__(*args, **kwargs)
+        self.fiberTraceProfileFittingControl = drpStella.FiberTraceProfileFittingControl()
```
ziggyman: moved initialization to `__init__` as well

python/pfs/drp/stella/createFlatFiberTraceProfileTask.py
```diff
@@ -51,20 +51,24 @@ class CreateFlatFiberTraceProfileTask(Task):
...
+
+    def getFiberTraceProfileFittingControl(self):
+        return self.fiberTraceProfileFittingControl
```
ziggyman: Removed getter function

python/pfs/drp/stella/createFlatFiberTraceProfileTask.py
```diff
@@ -76,16 +76,6 @@ class CreateFlatFiberTraceProfileTask(Task):
...
-#        self.makeSubtask("isr")
-#        self.schema = afwTable.SourceTable.makeMinimalSchema()
-#        self.makeSubtask("detection", schema=self.schema)
-#        self.makeSubtask("measurement", schema=self.schema)
-#        self.starSelector = self.config.starSelector.apply()
-#        self.candidateKey = self.schema.addField(
-#            "calib.psf.candidate", type="Flag",
-#            doc=("Flag set if the source was a candidate for PSF determination, "
-#                 "as determined by the '%s' star selector.") % self.config.starSelector.name
-#        )
```
ziggyman: done

python/pfs/drp/stella/createFlatFiberTraceProfileTask.py
```diff
@@ -82,8 +82,7 @@ def run(self, inFiberTraceSet, inTraceNumbers=[-1]):
 
         This method changes the input FiberTraceSet and returns void
         """
-        
+
         spectrumSet = self.createFlatFiberTraceProfile(inFiberTraceSet, inTraceNumbers)
-        
+
         return spectrumSet
-
```
ziggyman: done

python/pfs/drp/stella/findAndTraceAperturesTask.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, *args, **kwargs):
         super(FindAndTraceAperturesTask, self).__init__(*args, **kwargs)
         self.ftffc = drpStella.FiberTraceFunctionFindingControl()
 
+    def getFiberTraceFunctionFindingControl(self):
+        return self.ftffc
```
ziggyman: Removed

python/pfs/drp/stella/findAndTraceAperturesTask.py
```diff
@@ -65,9 +65,11 @@ class FindAndTraceAperturesTask(Task):
 
     def __init__(self, *args, **kwargs):
         super(FindAndTraceAperturesTask, self).__init__(*args, **kwargs)
+        self.ftffc = drpStella.FiberTraceFunctionFindingControl()
+
+    def findAndTraceApertures(self, inExposure):
```
ziggyman: done

python/pfs/drp/stella/findAndTraceAperturesTask.py
```diff
@@ -102,11 +102,10 @@ def run(self, inExposure):
...
+
         This method returns a FiberTraceSet
         """
-        
+
         fiberTraceSet = self.findAndTraceApertures(inExposure)
-        
+
         return fiberTraceSet
-
```
ziggyman: done

tests/Spectra.py
```diff
@@ -371,6 +369,49 @@ def testGetSpectra(self):
...
+
+        for interpolation in ['SPLINE3', 'PISKUNOV']:
+            myProfileTask.config.profileInterpolation = interpolation
+            myProfileTask.run(fts)
+
+            distanceFromCenterOrigProfRec = drpStellaMath.getDistTraceProfRec(fts)
+            dist, orig, rec = [distanceFromCenterOrigProfRec[:,0],
+                               distanceFromCenterOrigProfRec[:,1],
+                               distanceFromCenterOrigProfRec[:,3]]
+            ratio = orig / np.where(rec < 0.1, 1., rec)
```
ziggyman: Added comment why we do this. Essentially we avoid division by zero and multiplying the original pixel values with a large number. As there is no signal in these pixels, the SNR is gonna be small here anyway, so all those pixels are useless for flat-fielding

tests/Spectra.py
ziggyman: Well, what is "small"? The standard deviation is big where is a lot of signal (noise). This was the best test I could think of...

python/pfs/drp/stella/math.py
```diff
@@ -0,0 +1,119 @@
...
+    @brief Calculate the mean and standard deviations for bins in x
+    This function returns the mean and standard deviation of the y values in
+    bins in x with bin width <binWidthX>
+    @param x in: np.ndarray of x values
+           NOTE that the array must be sorted in X.
+    @param y in: np.ndarray of y values, same shape as x
+    @param binWidthX in: width of center distance bins
+    @return array of shape[nBins,2,2] containing:
+            [:,0,:]: the X for start and end of the bin,
+            [:,1,:]: the mean and standard deviation of the y values in each bin
```
ziggyman: Returning binRange, mean, and stdev separately now

tests/FiberTraces.py
ziggyman: Yes, `FiberTraceSet.createTraces` does change the `FiberTraceSet`. I offered this function so we can simply extract the FiberTraces from a new image without the need to copy all the object variables. I added a comment to `FiberTraces.h` to make this clear

PR #17, opened 2017-02-10
Tickets/pipe2 d 114

ziggyman:


include/pfs/drp/stella/SpectraTemplates.hpp
```diff
@@ -524,10 +523,16 @@ bool pfs::drp::stella::Spectrum<SpectrumT, MaskT, VarianceT, WavelengthT>::ident
...
     if ( math::isMonotonic( _wavelength ) == 0 ){
       cout << "Identify: WARNING: Wavelength solution is not monotonic => Setting identifyResult.rms to 1000" << endl;
+      cout << "Identify: _wavelength = " << _wavelength << endl;
       _dispRms = 1000.;
       LOGLS_WARN(_log, "Identify: RMS = " << _dispRms);
       LOGLS_WARN(_log, "======================================");
     }
+    else{
+      #ifdef __DEBUG_IDENTIFY__
+        cout << "identify: _wavelength is monotonic " << endl;
```
jdswinbank: Aren't we moving to using logging for this? Why are we adding more references to `cout`?

src/utils/Utils.cc
```diff
@@ -425,7 +425,7 @@ namespace pfs { namespace drp { namespace stella { namespace utils{
     double goodVal = yArr[badPos];
     yArr[badPos] = badVal;
 
-    /// Test PolyFit(xArr, yArr, nDeg)
+    cout << "Test PolyFit(xArr, yArr, nDeg)" << endl;
```
jdswinbank: I think the approach you are taking here is fundamentally wrong-headed.

You have a single giant function which tests a number of different calls to `PolyFit`. Rather than doing that, you ought to have a number of short unit tests, each of which tests a single call.

Ideally, each of the unit tests would be written in Python. However, I understand that the full `PolyFit` interface is not exposed to Python, so this is impractical. That leaves you with two choices:

- Use a C++-based test harness. The one in use in LSST is [Boost::Test](http://www.boost.org/doc/libs/1_63_0/libs/test/doc/html/index.html). For example, see [these tests in afw](https://github.com/lsst/afw/blob/a88e8e0d1ec5b1343fb138d9ed0cfbf5e923b889/tests/footprintArray.cc#L56).
- Write individual test functions in C++ then expose each one to Python by wrapping it separately using Swig.

Either of those approaches is adequate. What you've got now isn't.

However, I don't think you need to address that on this ticket. Therefore, I am minded to temporarily accept what you've done here, on the condition that you file another ticket to come back and clean up properly sometime in future.

include/pfs/drp/stella/Spectra.h
```diff
@@ -3,28 +3,13 @@
...
-//#pragma clang diagnostic ignored "-Wunused-variable"
-//#pragma clang diagnostic pop
-#include "boost/algorithm/string/trim.hpp"
-#include "lsst/base.h"
-#include "lsst/afw/geom.h"
-#include "lsst/afw/image/MaskedImage.h"
+#include "Controls.h"
 #include "lsst/log/Log.h"
-#include "lsst/pex/config.h"
 #include "lsst/pex/exceptions.h"
```
ziggyman: `pexExcept` IS being used in `SpectraTemplates.hpp`. The reason why removing it doesn't break the code is that it is also being included in `CurveFitting.h`

src/math/CurveFitting.cc
```diff
@@ -210,10 +189,8 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
                                                                                                         ndarray::makeVector(1))));
     LOGLS_DEBUG(_log, "P_D_A1_MeasureErrorsTemp = " << *P_D_A1_MeasureErrorsTemp);
 
-    for (size_t i = 0; i < I_A1_OrigPos.size(); ++i)
-      I_A1_OrigPos[i] = i;
     int I_NRejected_Old = 0;
-    std::vector<size_t> I_A1_Rejected_Old(D_A1_X_In.size());
+    std::vector<size_t> I_A1_Rejected_Old(0);
```
ziggyman: Before using `I_A1_Rejected_Old` it is resized to size 0, and then filled with `std::vector::push_bak`. I thought it is easier to read this way: declaring it with size 0 and reserving the space for the size of `D_A1_X_In`. Testing that this behavior is correct would require to rip this function apart into lots of small functions which could be tested individually. As all tests are passing I assume that this new behavior is correct.

src/math/CurveFitting.cc
```diff
@@ -304,31 +269,25 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
-        LOGLS_DEBUG(_log, "P_D_A1_MeasureErrorsTemp.getShape()[0] = " << P_D_A1_MeasureErrorsTemp->getShape()[0]);
       }
 
       B_Run = false;
       if (*P_I_NRejected != I_NRejected_Old)
         B_Run = true;
       else{
         for (int i_pos=0; i_pos < *P_I_NRejected; i_pos++){
-          if (std::fabs((*P_I_A1_Rejected)[i_pos] - I_A1_Rejected_Old[i_pos]) > 0.0001)
+          if (((*P_I_A1_Rejected)[i_pos] - I_A1_Rejected_Old[i_pos]) != 0)
```
ziggyman: As above. I actually changed it again to `if ((*P_I_A1_Rejected)[i_pos] != I_A1_Rejected_Old[i_pos])`. As both variables can only have integer values the new behavior must be the same as the old one, and all tests are passing, including the ones where we check the rejected values of PolyFit.

src/math/CurveFitting.cc
```diff
@@ -133,31 +114,28 @@ namespace pfs{ namespace drp{ namespace stella{ namespace math{
...
+        string message("pfs::drp::stella::math::CurveFitting::PolyFit: Error: P_D_A1_MeasureErrors->getShape()[0](=");
         message += to_string(P_D_A1_MeasureErrors->getShape()[0]) + ") != D_A1_X_In.getShape()[0](=" + to_string(D_A1_X_In.getShape()[0]) + ")";
         throw LSST_EXCEPT(pexExcept::Exception, message.c_str());
       }
     }
 
     PTR(std::vector<size_t>) P_I_A1_NotRejected(new std::vector<size_t>());
-    I_Pos = pfs::drp::stella::utils::KeyWord_Set(S_A1_Args_In, "NOT_REJECTED");
-    if (I_Pos >= 0){
+    if ((I_Pos = pfs::drp::stella::utils::KeyWord_Set(S_A1_Args_In, "NOT_REJECTED")) >= 0){
```
ziggyman: done

src/utils/Utils.cc
ziggyman: Created ticket PIPE2D-163

include/pfs/drp/stella/Controls.h
```diff
@@ -3,6 +3,8 @@
 
 #include <vector>
 #include "lsst/base.h"
+#include "lsst/pex/config.h"
+#include "ndarray.h"
```
ziggyman: The only explanation I have for some of the comments is that something went wrong during a rebase. I have fixed these commits again and compiled each one of them

include/pfs/drp/stella/Controls.h
```diff
@@ -13,7 +15,7 @@ namespace pfs { namespace drp { namespace stella {
  */
 struct FiberTraceFunctionControl {
   /// enum corresponding to legal values of interpolation string
-  enum INTERPOLATION {  POLYNOMIAL=0, CHEBYSHEV, NVALUES };
+  enum INTERPOLATION { POLYNOMIAL=0, CHEBYSHEV, NVALUES };
```
ziggyman: as above

include/pfs/drp/stella/Spectra.h
```diff
@@ -3,28 +3,13 @@
 
 #include <vector>
 #include <iostream>
-#include <cmath>
-//#pragma clang diagnostic push
-//#pragma clang diagnostic ignored "-Wunused-variable"
-//#pragma clang diagnostic pop
-#include "boost/algorithm/string/trim.hpp"
-#include "lsst/base.h"
-#include "lsst/afw/geom.h"
```
ziggyman: done

include/pfs/drp/stella/Spectra.h
```diff
@@ -3,28 +3,13 @@
...
-#include "lsst/afw/image/MaskedImage.h"
```
ziggyman: Added the include `lsst/afw/image/Image.h` which was only in a file included by `Spectra.h`

include/pfs/drp/stella/Spectra.h
```diff
@@ -3,28 +3,13 @@
...
-#include <fitsio.h>
-#include <fitsio2.h>
+#include "ndarray.h"
 #include "math/Math.h"
 #include "math/CurveFitting.h"
 #include "utils/Utils.h"
-#include "Controls.h"
-#include "lsst/afw/fits.h"
-#include "lsst/afw/image/fits/fits_io.h"
-#include "lsst/afw/image/fits/fits_io_mpl.h"
```
ziggyman: Re-ordered commits which fixed it

include/pfs/drp/stella/SpectraTemplates.hpp
ziggyman: Oups. fixed

include/pfs/drp/stella/math/CurveFitting.h
```diff
@@ -3,17 +3,13 @@
 
 #include <vector>
 #include <iostream>
-#include "lsst/base.h"
-#include "lsst/afw/geom.h"
```
ziggyman: removed

include/pfs/drp/stella/math/CurveFitting.h
```diff
@@ -3,17 +3,13 @@
...
-#include "lsst/afw/image/MaskedImage.h"
```
ziggyman: removed

include/pfs/drp/stella/math/CurveFitting.h
```diff
@@ -3,17 +3,13 @@
...
+#include "CurveFittingGaussian.h"
 #include "lsst/log/Log.h"
-#include "lsst/pex/config.h"
 #include "lsst/pex/exceptions/Exception.h"
```
ziggyman: we do need it

src/Spectra.cc
```diff
@@ -1,5 +1,4 @@
 #include "pfs/drp/stella/Spectra.h"
-#include "lsst/log/Log.h"
```
ziggyman: rebasing to master fixed that commit

PR #19, opened 2017-02-13
Tickets/pipe2 d 65

ziggyman:


src/Spectra.cc
```diff
@@ -945,7 +942,7 @@ namespace pfs { namespace drp { namespace stella { namespace math {
                                                                                                DispCorControl const& );
 
 }}}}
-
+
```
RobertLuptonTheGood: Please remove all pure white-space changes

tests/FiberTraces.py
```diff
@@ -854,10 +794,12 @@ def suite():
 
     suites = []
     suites += unittest.makeSuite(FiberTraceTestCase)
-#    suites += unittest.makeSuite(tests.MemoryTestCase)
     return unittest.TestSuite(suites)
 
 def run(exit = False):
+    """Quiet down loggers which are too verbose"""
+    lsst.log.setLevel("CameraMapper", lsst.log.FATAL)
+    lsst.log.setLevel("afw.image.ExposureInfo", lsst.log.FATAL)
```
RobertLuptonTheGood: I'm a little suspicious of this one.   What are the unwanted messages, and did you file a ticket upstream to get rid of them?

In general we should not be setting log levels for other packages.

tests/Spectra.py
```diff
@@ -451,6 +453,10 @@ def testWavelengthCalibrationWithRefSpec(self):
 
     def testWavelengthCalibrationWithoutRefSpec(self):
         if True:
+            logger = log.Log.getLogger("findAndTraceApertures")
+            logger.setLevel(log.WARN)
```
RobertLuptonTheGood: Why is this here, not in `setUpClass()`?

src/Spectra.cc
ziggyman: done

tests/FiberTraces.py
ziggyman: These are the outputs:

INFO  CameraMapper: Loading registry registry from /Users/azuri/stella-git/drp_stella_data/tests/data/PFS/registry.sqlite3
INFO  CameraMapper: Loading calibRegistry registry from /Users/azuri/stella-git/drp_stella_data/tests/data/PFS/CALIB/calibRegistry.sqlite3
INFO  afw.image.ExposureInfo: Empty WCS extension, using FITS header
INFO  afw.image.ExposureInfo: Empty WCS extension, using FITS header

I have not filed a ticket upstream yet but can do

tests/Spectra.py
ziggyman: moved to setUp()

tests/FiberTraces.py
PaulPrice: No need. Logging has changed a lot since LSST 12.1.

RobertLuptonTheGood: Also, why did you set the level to FATAL?  Was e.g. WARN not sufficient.

Paul: what are you proposing we do now?

PaulPrice: Upgrade our LSST code first, then re-evaluate the need for changing log levels. In the mean time, either live with the noise or suppress it, depending on how long you expect we're going to live in the current state.

PaulPrice: Let's leave this as it is until we upgrade LSST.

PR #22, opened 2017-03-10
Tickets/pipe2 d 53

ziggyman:


examples/runPipeline
```diff
@@ -1,3 +1,12 @@
+#!/bin/bash
+
+# On OS X El Capitan we need to pass through the library load path
```
jdswinbank: El Capitan *and later*. Plus it's now officially "macOS", not "OS X". :-)

PR #23, opened 2017-03-13
tests/testDRP.py: replace Popen with call to check the return value

ziggyman:
"subprocess.Popen" simply starts the command in the background while
the rest of the script is continued. In contrast to that,
"subprocess.call" waits for the command to complete and returns the
"returncode" attribute which we need to check to see if our test
was successful.

### DIFF
--- a/python/pfs/drp/stella/reduceArcRefSpecTask.py
+++ b/python/pfs/drp/stella/reduceArcRefSpecTask.py
@@ -43,7 +43,7 @@ class ReduceArcRefSpecTaskRunner(TaskRunner):
             try:
                 result = task.run(**args)
             except Exception, e:
-                task.log.fatal("Failed: %s" % e)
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/reduceArcRefSpecTask.py
```diff
@@ -42,6 +44,8 @@ def __call__(self, args):
                 result = task.run(**args)
             except Exception, e:
                 task.log.fatal("Failed: %s" % e)
+                traceback.print_exc(file=sys.stderr)
```
RobertLuptonTheGood: Why not just throw the exception?  Can't the user can use `--doraise` if desired?

ziggyman: Changed testDRP.py to use `--doraise`

PR #28, opened 2017-04-04
Add functions to read WavelengthFile, LineListFile, ReferenceSpectrum

ziggyman:
and writePfsArm,
include these functions in reduceArc(RefSpec)Task.py

### DIFF
--- a/python/pfs/drp/stella/reduceArcRefSpecTask.py
+++ b/python/pfs/drp/stella/reduceArcRefSpecTask.py
@@ -1,6 +1,5 @@
 import os
 
-from astropy.io import fits as pyfits
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -8,9 +7,9 @@ from lsst.pex.config import Config, Field
 from lsst.pipe.base import Struct, TaskRunner, ArgumentParser, CmdLineTask
 from lsst.utils import getPackageDir
 import pfs.drp.stella as drpStella
-from pfs.drp.stella.datamodelIO import spectrumSetToPfsArm, PfsArmIO
 import pfs.drp.stella.extractSpectraTask as esTask
-from pfs.drp.stella.utils import makeFiberTraceSet
+from pfs.drp.stella.utils import makeFiberTraceSet, readLineListFile
+from pfs.drp.stella.utils import readReferenceSpectrum, writePfsArm
 
 class ReduceArcRefSpecConfig(Config):
     """Configuration for reducing arc images"""
@@ -82,6 +81,13 @@ class ReduceArcRefSpecTask(CmdLineTask):
         if len(expRefList) == 0:
             raise RuntimeError("Unable to find exposure reference")
 
+        # read line list
+        lineListArr = readLineListFile(lineList)
+
+        # read reference Spectrum
+        refSpecArr = readReferenceSpectrum(refSpec)
+        self.log.debug('len(refSpecArr) = %d' % len(refSpecArr))
+
         for arcRef in expRefList:
             self.log.debug('arcRef.dataId = %s' % arcRef.dataId)
             self.log.debug('arcRef = %s' % arcRef)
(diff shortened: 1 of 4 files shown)


python/pfs/drp/stella/utils.py
```diff
@@ -128,3 +131,60 @@ def createPfsFiberTrace(dataId, fiberTraceSet, nRows):
...
+    traceIdsTemp = np.ndarray(shape=(len(tbdata)), dtype='int')
+    xCenters = np.ndarray(shape=(len(tbdata)), dtype='float32')
+    yCenters = np.ndarray(shape=(len(tbdata)), dtype='float32')
+    wavelengths = np.ndarray(shape=(len(tbdata)), dtype='float32')
+    traceIdsTemp[:] = tbdata[:]['fiberNum']
+    traceIds = traceIdsTemp.astype('int32')
+    wavelengths[:] = tbdata[:]['pixelWave']
+    xCenters[:] = tbdata[:]['xc']
+    yCenters[:] = tbdata[:]['yc']
+    return {'xCenters': xCenters, 'yCenters': yCenters, 'wavelengths': wavelengths, 'traceIds': traceIds}
```
jdswinbank: Why are you returning a `dict` here? I think any of `pipeBase.struct`, `collections.NamedTuple` or just a plain old `tuple` would be more idiomatic.

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -76,6 +75,18 @@ def run(self, expRefList, butler, wLenFile=None, lineList=None, immediate=True):
         self.log.info('wLenFile = %s' % wLenFile)
         self.log.info('lineList = %s' % lineList)
 
+        """ read wavelength file """
+        wavelengthFile = readWavelengthFile(wLenFile)
+        traceIds = wavelengthFile['traceIds']
+        wavelengths = wavelengthFile['wavelengths']
+        xCenters = wavelengthFile['xCenters']
+        yCenters = wavelengthFile['yCenters']
+        traceIdsUnique = np.unique(traceIds)
```
jdswinbank: This unpacking is tiresome. I comment on this below, but you could write this much more concisely if `readWavelengthFile` returned a `NamedTuple`.

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -76,6 +75,18 @@ def run(self, expRefList, butler, wLenFile=None, lineList=None, immediate=True):
...
+        self.log.debug('traceIdsUnique = %s' % (np.array_str(traceIdsUnique)))
+
+        """ read line list """
+        lineListArr = readLineListFile(lineList)
```
jdswinbank: Again moving outside the loop — same consideration as earlier. Great if it's safe!

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -149,7 +138,6 @@ def run(self, expRefList, butler, wLenFile=None, lineList=None, immediate=True):
                 spec.setITrace(iTraces[i])
                 self.log.info('flatFiberTraceSet.getFiberTrace(%d).getITrace() = %d, spec.getITrace() = %d' %(i,flatFiberTraceSet.getFiberTrace(i).getITrace(), spec.getITrace()))
                 specSpec = spec.getSpectrum()
-                self.log.info('yCenters.shape = %d' % yCenters.shape)
```
jdswinbank: Why? I think the log message was probably redundant (I guess it logged the same thing for every loop iteration), but it now seems to be gone entirely (ie not even logged once) without explanation. If you just don't want it any more, I'd have this on a separate commit with a message like "Removed redundant log message".

python/pfs/drp/stella/reduceArcRefSpecTask.py
```diff
@@ -78,6 +77,13 @@ def run(self, expRefList, butler, refSpec=None, lineList=None, immediate=True):
         self.log.debug('refSpec = %s' % refSpec)
         self.log.debug('lineList = %s' % lineList)
 
+        """ read line list """
+        lineListArr = readLineListFile(lineList)
+
+        """ read reference Spectrum """
+        refSpecArr = readReferenceSpectrum(refSpec)
+        self.log.debug('len(refSpecArr) = %d' % len(refSpecArr))
+
```
ziggyman: I agree. refSpecArr does not get changed

python/pfs/drp/stella/utils.py
ziggyman: changed readWavelengthFile to return a tuple and then
{{xCenters, wavelengths, traceIds = readWavelengthFile(self.wLenFile)}}

python/pfs/drp/stella/reduceArcTask.py
ziggyman: It is safe

python/pfs/drp/stella/reduceArcTask.py
ziggyman: see above comment

python/pfs/drp/stella/reduceArcTask.py
ziggyman: This line doesn't exist anymore due to previous code changes

PR #31, opened 2017-04-07
stellaLib.i: Remove where function, leave 1 for the tests

ziggyman:


ziggyman:
Nothing left to do after previous merges

PR #34, opened 2017-06-23
PIPE2D-172: Remove "if False:" blocks and get rid of "if True:"

jdswinbank:


tests/FiberTraces.py
```diff
@@ -125,528 +125,519 @@ def testFiberTraceFunctionFindingControl(self):
...
+            self.assertLess(np.absolute(stdDevDiff), 0.001)
+
+        # Fit profile with MkSlitFunc
+        ftpfc = drpStella.FiberTraceProfileFittingControl()
+        fiberTraceSet = drpStella.findAndTraceAperturesF(self.flat.getMaskedImage(), self.ftffc)
+        for iTrace in range(0, fiberTraceSet.size()):
+            fiberTrace = fiberTraceSet.getFiberTrace(iTrace)
             try:
-                fiberTraceMIF.setMask(maskedImageWrongSize.getMask())
+                spectrum = fiberTrace.calcProfile()
```
RobertLuptonTheGood: This looks like a substantive change, but maybe the diff is confused.

Are there any non-whitespace changes on this ticket?

tests/FiberTraces.py
```diff
@@ -125,528 +125,519 @@ def testFiberTraceFunctionFindingControl(self):
...
+            bool = fiberTrace.calcProfile()
+            self.assertTrue(bool)
 
-                spectrum = fiberTrace.extractFromProfile()
+            spectrum = fiberTrace.extractFromProfile()
 
-                recImage = fiberTrace.getReconstructed2DSpectrum(spectrum)
-                diff = fiberTrace.getTrace().getImage().getArray() - recImage.getArray()
-                if display:
-                    ds9.mtv(diff,title="reconstruction from MkSlitFunc",frame=fiberTraceSet.size()+iTrace)
```
RobertLuptonTheGood: There should be *no* ds9 calls.  Please always use `afwDisplay`; PIPE2D-211

Out of scope for this ticket, but please note.

tests/FiberTraces.py
ziggyman: Git diff really is confused here. I removed all {{if False:}} and {{if True:}} statements and adjusted the indentations

PR #39, opened 2017-07-28
Tickets/pipe2 d 233

ziggyman:


src/math/CurveFitting.cc
```diff
@@ -419,7 +419,9 @@ namespace pfs { namespace drp { namespace stella { namespace math {
     }
     LOGLS_DEBUG(_log, "xNew = " << xNew);
 
-    D_A1_SDevSquare.deep() = D_A1_MeasureErrors * D_A1_MeasureErrors;
+    auto itM = D_A1_MeasureErrors.begin();
+    for (auto it=D_A1_SDevSquare.begin(); it!=D_A1_SDevSquare.end(); ++it, ++itM)
+        *it = pow(static_cast<double>(*itM), 2.0);
```
RobertLuptonTheGood: {} please, and ::pow

src/math/CurveFitting.cc
```diff
@@ -510,28 +514,26 @@ namespace pfs { namespace drp { namespace stella { namespace math {
       else{
         i = 0;
       }
-      for (j = i; j <= I_Degree_In; j++){
-        (*P_D_A2_Covar)[ ndarray::makeVector( j, p-j ) ] = D_Sum;
+      for (j = i; j <= static_cast<int>(I_Degree_In); j++){
+        covarTemp[ ndarray::makeVector( j, p-j ) ] = D_Sum;
```
RobertLuptonTheGood: Does I_Degree_In need to be size_t?  int would be more natural, or unpack it into a const int at the top of the function.

src/math/CurveFitting.cc
ziggyman: Done

src/math/CurveFitting.cc
ziggyman: unpacked into const int at the top of the function

PR #47, opened 2018-10-03
PIPE2D-290: Fix normalisation of fibers in flats

PaulPrice:
And some cleanup.


hassanxp:
I discuss this with Craig today and he has no objections with the changes you've made. So all changes approved.
We can see following discussions during Paul Price's visit to Princeton end Oct if additional changes need to be made.

PR #54, opened 2019-02-07
PIPE2D-338: add makeLamConfig.py

PaulPrice:
Makes a PfsConfig for a LAM exposure, given the colors of the fibers
used.

### DIFF
--- a/bin.src/makeLamDesign.py
+++ b/bin.src/makeLamDesign.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+
+"""
+Executable script to create a PfiDesign for a LAM exposure, given the colors
+of fibers used.
+"""
+
+import numpy as np
+from pfs.datamodel import PfiDesign, TargetType
+
+# Mapping of colors to fiberIds
+# Constructed from a snippet by Fabrice Madec, "dummy cable B fibers"
+# https://sumire-pfs.slack.com/files/U3MLENNHH/FFS6P4UR5/dummy_cable_b_fibers.txt
+FIBER_COLORS = {"red1": [2],
+                "red2": [3],
+                "red3": [308],
+                "red4": [339],
+                "red5": [340],
+                "red6": [342],
+                "red7": [649],
+                "red8": [650],
+                "orange": [12, 60, 110, 161, 210, 259, 341],
+                "blue": [32, 111, 223, 289, 418, 518, 620],
+                "green": [63, 192, 255, 401, 464, 525, 587],
+                "yellow": [347, 400, 449, 545, 593, 641],
+                }
+
+# Mapping of colors to hash value
+# This scheme makes the hash look like binary, with a 1 if the color was used and 0 if not
+HASH_COLORS = {color: 16**ii for ii, color in enumerate(sorted(FIBER_COLORS.keys()))}
+
+
+def makePfiDesign(pfiDesignId, fiberId):
+    """Build a ``PfsConfig``
+
+    Parameters
+    ----------
+    pfiDesignId : `int`
+        Identifier for the top-end design. For our purposes, this is just a
+        unique integer.
+    fiberId : `numpy.ndarray` of `int`
+        Array of identifiers for fibers that will be lit.
+
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/reduceExposure.py
```diff
@@ -199,16 +199,18 @@ def run(self, sensorRefList):
 
         fiberTraceList = [sensorRef.get("fibertrace") for sensorRef in sensorRefList]
         detectorMapList = [sensorRef.get("detectormap") for sensorRef in sensorRefList]
-        pfsConfig = sensorRefList[0].get("pfsConfig")
+        pfsConfig = None
```
RobertLuptonTheGood: Why is this part of PIPE2D-338?

python/pfs/drp/stella/reduceExposure.py
```diff
@@ -199,16 +199,18 @@ def run(self, sensorRefList):
...
 
         results.fiberTraceList = fiberTraceList
         results.detectorMapList = detectorMapList
-        results.pfsConfig = pfsConfig
 
         if self.config.doSubtractSky2d:
+            pfsConfig = sensorRefList[0].get("pfsConfig")
```
RobertLuptonTheGood: Don't we need the pfsConfig whenever we are extracting spectra or solving for the wavelength solution?  It's the only way that we know which fibres are illuminated.   Yes, we can guess but now that we have pfsConfigs from LAM I think we should expect to use them.

bin.src/makeLamDesign.py
```diff
@@ -0,0 +1,116 @@
...
+                "red": [2, 3, 308, 339, 340, 342, 649, 650],
+                "orange": [12, 60, 110, 161, 210, 259, 341],
+                "blue": [32, 111, 223, 289, 418, 518, 620],
+                "green": [63, 192, 255, 401, 464, 525, 587],
+                "yellow": [347, 400, 449, 545, 593, 641],
+                }
+
+# Mapping of colors to hash value
+# This scheme makes the hash look like binary, with a 1 if the color was used and 0 if not
+HASH_COLORS = {color: 16**ii for ii, color in enumerate(sorted(FIBER_COLORS.keys()))}
```
RobertLuptonTheGood: Maybe not worth changing at this point, but I think bit manipulations are better expressed as such (in this case by using `<<`)

bin.src/makeLamDesign.py
```diff
@@ -0,0 +1,116 @@
...
+    -------
+    hash : `int`
+        Hash, for the pfiDesignId.
+    """
+    return sum(HASH_COLORS[col] for col in set(colors))
+
+
+if __name__ == "__main__":
+    import argparse
+    parser = argparse.ArgumentParser(description="Create a PfiDesign for a LAM exposure, "
```
RobertLuptonTheGood: I think we should explain what "colors" [sic] of fibres means.  It's very LAM specific, and while I understand that this is LAM specific code it's in the PFS codebase

python/pfs/drp/stella/reduceExposure.py
PaulPrice: It deals with the original problem in a different way. Instead of constructing a `pfsConfig` for the data, this realises that we didn't need it in the first place (for some operations).

python/pfs/drp/stella/reduceExposure.py
PaulPrice: The `pfsConfig` tells you what the fiber is pointing at. For processing arcs, all the fibers are pointing at the same thing. The difference comes when we start doing odd/even exposures, but the pipeline will need to have a slight redesign to deal with that.
Currently, the way we know which fibers are illuminated is through the fiberTrace.

CraigLoomis: But it also tells you which fibers are illuminated, so the findTraces procedure would not have to guess (and it is a difficult guess to make for such spare "pluggings")? I think it is important to be able to trust the fiberTrace list.

PaulPrice: Sure, but that involves getting the `pfsConfig` into `constructFiberTrace.py`, which is a different feature.

PR #55, opened 2019-02-11
PIPE2D-316: Demonstrate arm merging

PaulPrice:


src/DetectorMap.cc
```diff
@@ -379,7 +367,7 @@ DetectorMap::_set_xToFiberId()
     float lastCenter = -1e30;            // center of the last fibre we passed scanning to the right
     std::size_t lastIndex = -1;          // index of the last fibre we passed.  Special case starting with 0
 
-    float nextCenter = getXCenter(_fiberIds[1], iy); // center of the next fiber we're looking at
```
RobertLuptonTheGood: Are you sure about this?  fiberIds start at 1, not 0.

src/DetectorMap.cc
```diff
@@ -48,18 +48,6 @@ DetectorMap::DetectorMap(
                                                           wavelengthKnots[ii].end());
         float const maxWavelengthKnot = *std::max_element(wavelengthKnots[ii].begin(),
                                                           wavelengthKnots[ii].end());
-        if (minCenterKnot < bbox.getMinY() || maxCenterKnot > bbox.getMaxY()) {
```
RobertLuptonTheGood: When do we need to allow this?

src/DetectorMap.cc
PaulPrice: This isn't about the fiberIds, but the indices of the vector, which start at zero.

src/DetectorMap.cc
PaulPrice: The Simulator produces knots that lie off the detector. I could imagine that if they are not included, the solution around the edge could be bad.

src/DetectorMap.cc
RobertLuptonTheGood: I know that, but I was concerned that I might have adjusted the size of that array.  If you've checked that I didn't, this is fine.

PaulPrice: I just double-checked, and I don't see any size adjustment or sneaky indexing going on. Perhaps there was, once upon a time (?), but there isn't now.

PR #57, opened 2019-02-20
PIPE2D-340: Provide method to adjust detector map slit offsets

PaulPrice:


tests/test_DetectorMap.py
```diff
@@ -329,6 +327,159 @@ def testPickle(self):
...
+    def testY(self):
+        """Slit offset in y"""
+        self.slitOffsets[drpStella.DetectorMap.DY][:] = -5.0
+        self.detMap.setSlitOffsets(self.slitOffsets)
+        self.assertPositions()
+
+    def testFocus(self):
+        """Offset focus
+
+        Doesn't do anything yet.
```
hassanxp: Please change to FIXME. We need to file a new ticket to implement the change to a detectormap when the focus changes.

PR #60, opened 2019-03-27
replaced references to expId with visit

hassanxp:


hassanxp:
Close pull request. Replacing expId with visit requires further discussion and clarification.

PR #61, opened 2019-04-01
PIPE2D-343: Design PSF class

PaulPrice:


python/pfs/drp/stella/psf.py
```diff
@@ -0,0 +1,251 @@
...
+        kernel = self.computeKernelImageXY(point)
+        xWhole, xFrac = positionToIndex(point.getX(), True)
+        yWhole, yFrac = positionToIndex(point.getY(), True)
+        if xFrac != 0.0 or yFrac != 0.0:
+            image = offsetImage(kernel, xFrac, yFrac, self.warpAlgorithm, self.warpBuffer)
+        image.setXY0(xWhole + image.getX0(), yWhole + image.getY0())
+        return image
+
+    def computeKernelImage(self, fiberId, wavelength):
+        """Return an image with the PSF centered at 0,0 (when xy0 is respected)
```
hassanxp: What is meant by 'when xy0 is respected'?

PaulPrice:
Fixed.

python/pfs/drp/stella/psf.py
PaulPrice: The `lsst.afw.image.Image` class (and relatives) use two coordinate systems:
* `LOCAL`: this is the coordinate system of the image in memory. The lower-left pixel is `0,0`.
* `PARENT`: this is a logical coordinate system that is offset from `LOCAL` by integer values `x0,y0` (often abbreviated to `xy0`).

So the above statement means that the PSF may not be centered at the lower-left corner (`0,0` in the `LOCAL` frame), but the location of the center of the PSF is `0,0` in the `PARENT` frame.
This is the same language used in [`lsst::afw::detection::Psf::computeKernelImage`](http://doxygen.lsst.codes/stack/doxygen/x_masterDoxyDoc/classlsst_1_1afw_1_1detection_1_1_psf.html#ab6d8cec1248cb14247bd9d897f13597b).

PR #62, opened 2019-04-03
PIPE2D-342: add definition of Lsf

PaulPrice:
Lsf is an abstract base class based on LSST's Psf class, but
brought down to one dimension.

This includes a Kernel class (for convolving by your Lsf, or
inserting the Lsf at a particular point in the spectrum) and
a couple of basic concrete Lsf classes for demonstration.

### DIFF
--- a/python/pfs/drp/stella/lsf.py
+++ b/python/pfs/drp/stella/lsf.py
@@ -1,0 +1,693 @@
+from abc import ABC, abstractmethod
+from functools import partial
+import pickle
+import numpy as np
+import scipy
+from scipy.interpolate import interp1d
+from scipy.integrate import trapz
+
+from lsst.geom import Point2D, Point2I
+from lsst.afw.image import ImageF, ImageD
+from lsst.afw.math import FixedKernel
+from lsst.afw.geom.ellipses import Quadrupole
+
+__all__ = ("Kernel1D", "Lsf", "GaussianKernel1D", "GaussianLsf", "FixedEmpiricalLsf")
+
+# Default interpolator factory
+DEFAULT_INTERPOLATOR = partial(interp1d, kind="linear", bounds_error=False, fill_value=0, copy=True,
+                               assume_sorted=True)
+
+
+class Kernel1D:
+    """A one-dimensional kernel
+
+    A ``Kernel1D`` is essentially a one-dimensional array with an indexing
+    offset, so that index 0 refers to the center of the kernel. Main operations
+    are ``convolve`` to convolve an array by a kernel, and ``toArray`` which
+    inserts the kernel at any point in an array.
+
+    Parameters
+    ----------
+    values : `numpy.ndarray`, floating-point, shape ``(N,)``
+        Array of kernel values.
+    center : `int`
+        Index of center of kernel in the ``values``. It needn't be the middle.
+    normalize : `bool`, optional
+        Normalize the kernel to unity?
+    """
+    def __init__(self, values, center, normalize=True):
+        self.length = len(values)
+        self.center = center
+        self.values = values
+        self.indices = np.arange(self.length) - center
+        self.min = self.indices[0]
+        self.max = self.indices[-1]
+        if normalize:
+            self.values /= self.normalization()
+
+    @classmethod
+    def makeEmpty(cls, halfSize):
+        """Create an empty ``Kernel1D``
+
+        Parameters
+        ----------
+        halfSize : `int`
+            Half-size of the kernel.
+
+        Returns
+        -------
+        self : ``cls``
+            Empty kernel.
+        """
+        size = 2*halfSize + 1
+        return cls(np.zeros(size, dtype=float), halfSize, normalize=False)
+
+    def __len__(self):
+        """Number of elements in kernel array"""
+        return self.length
+
+    def __getitem__(self, index):
+        """Get value(s) from the kernel
+
+        Supports integer or array indexing.
+
+        Values off the end are zero.
+        """
+        if isinstance(index, int):
+            return self.values[index + self.center]
+        result = np.empty_like(index)
+        rr, ii, vv = np.broadcast_arrays(result, index, self.values)
+        rr[:] = vv[ii + self.center]
+        return result
+
+    def __setitem__(self, index, value):
+        """Set value(s) in the kernel.
+
+        Supports integer or array indexing.
+
+        Values off the end raise an `IndexError`.
+        """
+        ii, vv = np.broadcast_arrays(index, value)
+        self.values[ii + self.center] = vv
+
+    def __iter__(self):
+        """Iterator
+
+        Produces ``index,value`` pairs.
+        """
+        return zip(iter(self.indices), iter(self.values))
+
+    def _getOtherValues(self, other):
+        """Get values from another Kernel1D IFF dimensions match"""
+        if isinstance(other, Kernel1D):
+            if self.min != other.min or self.max != other.max:
+                raise IndexError("Kernel1D ranges don't match: %d..%d vs %d..%d" %
+                                 (self.min, self.max, other.min, other.max))
+            other = other.values
+        return other
+
+    def __eq__(self, other):
+        """Test for equality"""
+        return (self.__class__ == other.__class__ and
+                self.min == other.min and
+                self.max == other.max and
+                np.all(self.values == other.values))
+
+    def __imul__(self, other):
+        """Multiplication in-place
+
+        Supports multiplication by a scalar or an array.
+        """
+        self.values *= self._getOtherValues(other)
+        return self
+
+    def __itruediv__(self, other):
+        """Division in-place
+
+        Supports division by a scalar or an array.
+        """
+        self.values /= self._getOtherValues(other)
+        return self
+
+    def __reduce__(self):
+        """Pickle prescription"""
+        return self.__class__, (self.values, self.center, False)
+
+    def __repr__(self):
+        """Representation"""
+        return "%s(%s, %f)" % (self.__class__.__name__, self.values, self.center)
+
+    def normalization(self):
+        """Return the normalization"""
+        return np.sum(self.values)
+
+    def convolve(self, array):
+        """Convolve an array by the kernel
+
+        The convolved array has the same length as the input array in the usual
+        case that the array is longer than the kernel.
+
+        Note that the ends of the array, the kernel does not overlap completely,
+        and there will be boundary effects (implementation is ``numpy.convolve``
+        with ``mode="same"``).
+
+        Parameters
+        ----------
+        array : `numpy.ndarray`, floating-point, shape ``(M,)``
+            Array to convolve.
+
+        Raises
+        ------
+        RuntimeError
+            If the array is shorter than the kernel.
+        """
+        if len(array) < self.length:
+            raise RuntimeError("Array too short for convolution by this kernel")
+        if self.center != (self.length - 1)//2:
+            # Pad to put center of kernel in the middle, so we don't introduce a shift
+            left = self.center
+            right = self.length - self.center - 1
+            middle = max(left, right)
+            length = 2*middle + 1
+            kernel = np.zeros(length, dtype=self.values.dtype)
+            kernel[middle - left:middle + right + 1] = self.values
+        else:
+            kernel = self.values
+        return np.convolve(array, kernel, "same")
+
+    def toArray(self, length, center, interpolator=DEFAULT_INTERPOLATOR):
+        """Convert to an array at a nominated position
+
+        Parameters
+        ----------
+        length : `int`
+            Desired length of array.
+        center : `float`
+            Index position at which to insert the kernel.
+        interpolator : callable, optional
+            Interpolator factory. Calling this with arrays of ``x`` and ``y``
+            should produce an interpolator that can be called with an array of
+            ``x`` values at which to interpolate.
+
+        Returns
+        -------
+        array : `numpy.ndarray`, floating-point, shape ``(length,)``
+            Array containing inserted kernel.
+        """
+        return interpolator(self.indices + center, self.values)(np.arange(length))
+
+    def computeStdev(self):
+        """Compute standard deviation of kernel"""
+        centroid = np.sum((self.indices*self.values).astype(np.float64))
+        return np.sqrt(np.sum((self.values*(self.indices - centroid)**2).astype(np.float64)))
+
+    def interpolate(self, indices):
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        return index < self.min or index > self.max
+
+    def _getitem(self, index):
+        """Get a single value from the kernel
+
+        Values off the end are zero.
+        """
+        return self.values[index + self.center] if not self._indexOutOfRange(index) else 0.0
+
+    def _setitem(self, index, value):
```
RobertLuptonTheGood: Isn't this a `__` method, solely used by `__setitem__`?
And again, aren't we doing the index check twice, once here, then in the index operation?  I assume that you're trying to provide better error messages, but we could do that in a `except` block

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        """Return the normalization"""
+        return np.sum(self.values)
+
+    def convolve(self, array):
+        """Convolve an array by the kernel
+
+        The convolved array has the same length as the input array in the usual
+        case that the array is longer than the kernel.
+
+        Note that the ends of the array will not be fully convolved.
```
RobertLuptonTheGood: What exactly does this mean?  I understand end effects, but I think you should be more explicit about what we are doing for people who don't know, or don't look at your code

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+
+        Raises
+        ------
+        RuntimeError
+            If the array is shorter than the kernel.
+        """
+        if len(array) < self.length:
+            raise RuntimeError("Array too short for convolution by this kernel")
+        if self.center != (self.length - 1)//2:
+            # Pad to put center of kernel in the middle, so we don't introduce a shift
```
RobertLuptonTheGood: A gross shift.  If the kernel has a non-zero first moment around the centre it will still shift and that will be Very Bad.  Probably should check in the ctor

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+            middle = max(left, right)
+            length = 2*middle + 1
+            kernel = np.zeros(length, dtype=self.values.dtype)
+            kernel[middle - left:middle + right + 1] = self.values
+        else:
+            kernel = self.values
+        return np.convolve(array, kernel, "same")
+
+    def toArray(self, length, center, interpolator=DEFAULT_INTERPOLATOR):
+        """Convert to an array at a nominated position
```
RobertLuptonTheGood: "convert to an array" implies just a reformatting, but you're doing some sort of interpolation.  What's actually happening here?

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+
+        Returns
+        -------
+        array : `numpy.ndarray`, floating-point
+            Array containing inserted kernel.
+        """
+        return interpolator(self.indices + center, self.values)(np.arange(length))
+
+    def computeStdev(self):
+        """Compute standard deviation of kernel"""
```
RobertLuptonTheGood: You mean "second moment" don't you?  Assuming that the first moment vanishes

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        centroid = np.sum((self.indices*self.values).astype(np.float64))
+        return np.sqrt(np.sum((self.values*(self.indices - centroid)**2).astype(np.float64)))
+
+    def interpolate(self, indices):
```
RobertLuptonTheGood: What is the use case for this? Is it intended to provide an interface to properly over-sampled (and maybe Zernike-based) models?

In other words, is `Kernel1D` to be thought of as a particular, rather over-simplified, kernel or are you thinking of it as a real working class?

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        """
+        return np.interp(indices, self.indices, self.values, left=0.0, right=0.0)
+
+
+class Lsf(ABC):
+    """Abstract base class for a Line-Spread Function
+
+    A line-spread function (LSF) is the one-dimensional version of a
+    point-spread function (PSF). It models the response of the spectrograph to a
+    single infinitely thin line, which may be a function of position.
```
RobertLuptonTheGood: I think we usually say, "narrow" for spectral lines

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+
+    This class includes two main styles of methods:
+
+    - methods taking a ``center`` position of type `float`. These methods are
+      intended for interacting with a numpy array.
+    - methods taking a ``point`` of type `lsst.geom.Point2D` (optionally; if not
+      provided, the ``point`` defaults to the result of ``getAveragePosition``).
+      These methods are  intended for emulating (duck-typing) an
+      `lsst.afw.detection.Psf` for when interacting with a
+      `lsst.afw.image.Image`.
```
RobertLuptonTheGood: Please provide an example (e.g. subtracting a spectrum specified as a set of (intensity, lambda) values)

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        if point is None:
+            point = self.getAveragePosition()
+        return ImageF(self.computeArray(point.getX())[np.newaxis, :].astype(np.float32))
+
+    def computeArray(self, center):
+        """Return an array with the LSF inserted at the nominated position
+
+        Besides the difference in return types from the ``computeKernel``
+        method (this returns an array, while ``computeKernel`` returns a
+        `Kernel1D`), this method allows positioning of the LSF at sub-pixel
```
RobertLuptonTheGood: I don't understand.  The image based version must handle sub-pixel offsets too (as does the `Psf` class which returns the sub-image with XY0 set properly)

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        Interpolator factory for kernel. Calling this with arrays of ``x`` and
+        ``y`` should produce an interpolator that can be called with an array of
+        ``x`` values at which to interpolate.
+    """
+    def __init__(self, length, interpolator=DEFAULT_INTERPOLATOR):
+        self.length = length
+        self.interpolator = interpolator
+
+    def computeImage(self, point=None):
+        """Return an image with the LSF inserted at the nominated position
```
RobertLuptonTheGood: I think this means, "return an image of the PSF at the nominated position"

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        kernel = self.computeKernel(point.getX())
+        return ImageF(kernel.values[np.newaxis, :].astype(np.float32), True, Point2I(-kernel.center, 0))
+
+    @abstractmethod
+    def computeKernel(self, center):
+        """Return a kernel for the nominated position
+
+        Besides the difference in return types from the ``computeArray``
+        method (this returns a `Kernel1D`, while ``computeArray`` returns a
+        an array), this method provides a model for the LSF centered at zero.
```
RobertLuptonTheGood: What does, "centred at zero" mean?  Not pixel `0`, but in the sense of your `Kernel1D` class?

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        """
+        raise NotImplementedError("Subclasses must implement this method")
+
+    def computePeak(self, point=None):
+        """Return the value of the peak pixel in the LSF image
+
+        Parameters
+        ----------
+        point : `lsst.geom.Point2D`, optional
+            Position at which to evaluate the LSF.
```
RobertLuptonTheGood: position in which coordinate system?

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        """
+        if point is None:
+            point = self.getAveragePosition()
+        sigma = self.computeShape1D(point.getX())
+        return Quadrupole(sigma**2, 0, 0)
+
+    def computeShape1D(self, center):
+        """Return the standard deviation at the nominated position
+
+        The standard deviation is a measure of the width of the LSF.
```
RobertLuptonTheGood: again, please refer consistently to "second moment"  (or `sqrt(I_xx)` if you want the sqrt and have defined `I_xx`)

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        """
+        return self.computeKernel(center).computeStdev()
+
+    def computeApertureFlux(self, radius, point=None):
+        """Return the flux within the nominated radius
+
+        Parameters
+        ----------
+        point : `lsst.geom.Point2D`, optional
+            Position at which to evaluate the LSF.
```
RobertLuptonTheGood: Define radius, presumably a `float`

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        -------
+        flux : `float`
+            Flux within the nominated radius.
+        """
+        if point is None:
+            point = self.getAveragePosition()
+        kernel = self.computeKernel(point.getX())
+        intRadius = int(radius)
+
+        flux = trapz([kernel[ii] for ii in range(-intRadius, intRadius + 1)])
```
RobertLuptonTheGood: How do you know that a trapezoidal integration is good enough?  Does the (much smarter but sampling-sensitive) SDSS code run?  If the LSF is oversampled (as in general it will be) we need to run on the fully-sampled version not on this instrument-sampled version

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+            Position at which to evaluate the LSF.
+
+        Returns
+        -------
+        bbox : `lsst.afw.geom.Box2I`
+            Bounding box for the kernel image.
+        """
+        if point is None:
+            point = self.getAveragePosition()
+        return self.computeKernelImage(point.getX()).getBBox()
```
RobertLuptonTheGood: That's pretty lazy!  It's probably OK, unless someone wants to use this method to implement `computeKernelImage` for a subclass

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+
+    @classmethod
+    def readFits(cls, filename):
+        """Read from file
+
+        Though the method name says "FITS", we read a pickle file, because
```
RobertLuptonTheGood: So this is a gen2 butler hack?

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        pickling takes care of subclasses for us.
```
RobertLuptonTheGood: Hmm.  We cannot write anything that we wish to keep as pickle files as they become unreadable if we change things in the implementation.  Probably OK in the short run, but it must be flagged.

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+        Parameters
+        ----------
+        filename : `str`
+            Filename to which to write.
+        """
+        with open(filename, "wb") as fd:
+            pickle.dump(self, fd)
+
+
+def gaussian(indices, sigma):
```
RobertLuptonTheGood: Why expose this?  And please don't call the width parameter `sigma` -- that implies statistics, not the convenient but inaccurate approximation of a PSF as a Gaussian

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+            Position at which to evaluate the LSF. Not that it matters, as it's
+            the same everywhere.
+
+        Returns
+        -------
+        kernel : `Kernel1D`
+            Kernel with a model of the LSF at the nominated position.
+        """
+        return self.kernel
+
```
RobertLuptonTheGood: Why don't you overload `computeApertureFlux`?

tests/test_lsf.py
```diff
@@ -0,0 +1,274 @@
...
+    Parameters
+    ----------
+    array : array-like
+        Array for which to calculate moments.
+
+    Returns
+    -------
+    centroid : `float`
+        First moment.
+    stdev : `float`
```
RobertLuptonTheGood: Not `stdev`!  `rms` would be OK

python/pfs/drp/stella/lsf.py
```diff
@@ -0,0 +1,688 @@
...
+    def _indexOutOfRange(self, index):
+        """Is the index out of range?"""
+        return index < self.min or index > self.max
+
+    def _getitem(self, index):
+        """Get a single value from the kernel
+
+        Values off the end are zero.
+        """
+        return self.values[index + self.center] if not self._indexOutOfRange(index) else 0.0
```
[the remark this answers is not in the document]
PaulPrice: I think it's helpful to the user if the kernel goes to zero outside the range where it's explicitly defined.

python/pfs/drp/stella/lsf.py
PaulPrice: Leading double underscore is [used to prevent the method from being overridden by subclasses](http://igorsobreira.com/2010/09/16/difference-between-one-underline-and-two-underlines-in-python.html), so I don't think it's appropriate here.
But I've reworked the getter and setter by using `np.broadcast_arrays` instead of `np.vectorize`; that should be faster, and it eliminates the confusing `_getitem` and `_setitem` methods.

python/pfs/drp/stella/lsf.py
PaulPrice: DAMD-78. I don't think it makes sense to schedule this until we've had a chance to play around with the `Lsf` classes here and get a feel for what the datamodel should be.

python/pfs/drp/stella/lsf.py
PaulPrice: I wouldn't have said it's exposed since it's not included in `__all__`.

python/pfs/drp/stella/lsf.py
PaulPrice: I don't remember if it's because I'm lazy or if I wanted to sum the pixels. In any case, I don't see it being used very much, and so I'm going to default to being lazy.

PR #65, opened 2019-05-16
PIPE2D-415: not finding all the fibers

PaulPrice:


hassanxp:
The changes all look fine to me.

PR #68, opened 2019-05-22
PIPE2D-427: reduceExposure: mask negative variance pixels

PaulPrice:
We have low count levels compared to imaging, so we sometimes have
pixels that are negative after bias+dark, which means the variance
gets set to a negative number, which causes problems downstream
(e.g., CurveFitting.cc:330). Find these pixels, set the variance
to infinity and mask them as BAD.

### DIFF
--- a/python/pfs/drp/stella/reduceExposure.py
+++ b/python/pfs/drp/stella/reduceExposure.py
@@ -20,6 +20,7 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 from collections import defaultdict
+import numpy as np
 
 from lsst.pex.config import Config, Field, ConfigurableField
 from lsst.pipe.base import CmdLineTask, TaskRunner, Struct
@@ -181,7 +182,7 @@ class ReduceExposureTask(CmdLineTask):
         psfList = []
         lsfList = []
         for sensorRef in sensorRefList:
-            exposure = self.isr.runDataRef(sensorRef).exposure
+            exposure = self.runIsr(sensorRef)
             if self.config.doRepair:
                 self.repairExposure(exposure)
             if self.config.doSkySwindle:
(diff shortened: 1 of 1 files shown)


hassanxp:
Why is the variance set to infinity? Isn't masking them as BAD sufficient?

python/pfs/drp/stella/reduceExposure.py
```diff
@@ -230,6 +231,28 @@ def run(self, sensorRefList):
         self.write(sensorRefList, results)
         return results
 
+    def runIsr(self, sensorRef):
+        """Run Instrument Signature Removal (ISR)
+
+        This method exists to allow us to post-process the ISR-ed image.
```
hassanxp: How about 'This method wraps the ISR call to allow us to post-process the ISR-ed image.'

python/pfs/drp/stella/reduceExposure.py
```diff
@@ -230,6 +231,28 @@ def run(self, sensorRefList):
...
+
+        Returns
+        -------
+        exposure : `lsst.afw.image.Exposure`
+            Sensor image after ISR has been applied.
+        """
+        exposure = self.isr.runDataRef(sensorRef).exposure
+        # Remove negative variance (because we can have very low count levels)
+        bad = np.where(exposure.variance.array < 0)
+        exposure.variance.array[bad] = np.inf
```
hassanxp: See my earlier comment about the rationale of assigning infinity to a variance value already assigned as bad. It's a good idea to choose a single value, but why not another nonsensical number such as -1? Would that value result in the same failure?

PaulPrice: The variance must be non-negative. An infinite variance gives the pixel no weight in calculations without any additional effort, even if the user isn't respecting the mask.

hassanxp: Understood.

RobertLuptonTheGood:
I don't think that this fix is correct.

The problem is that the variance is being calculated after bias and dark subtraction as the number of photons.  Firstly, the dark current must be included in the variance (maybe this was a typo?).  More importantly, the variance shouldn't be the measured number of photons but its expectation value.

Masking pixels that end up negative biases the image, and this is a problem.  I think that we won't need to do any masking;  if we do, we need to think harder about the consequences.

Even if the pixel must be taken as bad (which is not, I think, correct) please do not set the variance to infinity; that causes biases too.  You should interpolate the variance at the same time as interpolating the image (and setting BAD).

On a technical level, none of the working tasks should be accepting a ButlerDataRef;  please unpack the data types you need and pass them into the function.  The main task method should be called run

PaulPrice:
Some of this is because we're currently locked into LSST v16.0 and its ISR. I'll make a ticket to fix the variance.

PR #72, opened 2019-07-02
PIPE2D-436: Investigate wavelength solution quality

PaulPrice:


bin.src/makeLamDesign.py
```diff
@@ -23,6 +23,7 @@
                 "blue": [32, 111, 223, 289, 418, 518, 620],
                 "green": [63, 192, 255, 401, 464, 525, 587],
                 "yellow": [347, 400, 449, 545, 593, 641],
+                "xxx": [650, 589, 462, 400, 337, 315, 253, 191, 65, 3],  # Before there were colors
```
hassanxp: Can you explain a little more the purpose of this entry?

PaulPrice: The commit message adds some explanation:
> So we can process data from long ago, when we didn't
> have fiber colors.

These are the fibers that Craig determined were used in LAM data from 2017.

PR #97, opened 2020-03-02
PIPE2D-466: Move algorithmic code from datamodel

PaulPrice:


python/pfs/drp/datamodel/drp.py
```diff
@@ -0,0 +1,25 @@
+import pfs.datamodel.drp
+from .pfsSpectra import PfsSpectra
+from .pfsSpectrum import PfsSpectrum, PfsSimpleSpectrum
+
+__all__ = ("PfsArm", "PfsMerged", "PfsReference", "PfsSingle", "PfsObject")
+
+
+class PfsArm(pfs.datamodel.drp.PfsArm, PfsSpectra):
```
hassanxp: We have the class `pfs.datamodel.drp.PfsArm` from the datamodel repository, and the subclass `pfs.drp.datamodel.drp.PfsArm` in the drp_stella repository. This is confusing. Both module names contain 'datamodel' and the drp_stella module has 'drp' twice, with no hint as to which repo it is in. Could the module name be different? For example `pfs.drp.stelladm.dmodel.PfsArm` for the drp_stella `PfsArm` subclass?

python/pfs/drp/stella/SpectrumSetContinued.py
```diff
@@ -6,7 +6,7 @@
 from lsst.pipe.base import Struct
 import lsst.afw.image as afwImage
 
-from pfs.datamodel.drp import PfsArm
+from pfs.drp.datamodel import PfsArm
```
hassanxp: This illustrates the confusion with the naming of the drp_stella-specific classes: it's not obvious that with this change the drp_stella subclasses of the datamodel are being used.

PR #123, opened 2020-06-26
PIPE2D-611: reduceArc should also fix the x-direction of the DetectorMap

PaulPrice:


python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -28,7 +29,20 @@ class ReduceArcConfig(pexConfig.Config):
...
+        allowed={"NONE": "No tweaking",
+                 "ARC": "Tweak based on the arc lines",
+                 "FIBERTRACE": "Tweak based on the fiberTrace",
+                 },
+    )
+    arcTweakWidth = pexConfig.Field(dtype=int, default=5, doc="Half-width for centroiding in x")
+    arcTweakHeight = pexConfig.Field(dtype=int, default=2, doc="Half-height for stacking when centroiding")
+    arcTweakMask = pexConfig.ListField(dtype=str, default=["BAD", "SAT", "NO_DATA"],
+                                       doc="Mask planes to reject when centroiding arc lines")
+    arcTweakRms = pexConfig.Field(dtype=float, default=2.0, doc="Guess line RMS when centroiding arc lines")
```
hassanxp, line 45: Very minor comment: don't like the use of the word 'tweak' (or 'hack' in formal class/method/fn/parameter names). How about 'adjust'? Anyway, I won't object to that word being used if there's no better alternative or too much work to rename.

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -395,6 +409,75 @@ def updateCenters(self, fiberTraces, detectorMap):
...
+            Arc image.
+        spectra : `pfs.drp.stella.SpectrumSet`
+            Extracted spectra. We don't use the actual spectra, but this has
+            both the ``fiberId`` and reference line lists.
+        detectorMap : `pfs.drp.stella.DetectorMap`
+            Mapping of wl,fiber to detector position.
+        """
+        width, height = exposure.getDimensions()
+        badBitMask = exposure.mask.getPlaneBitMask(self.config.arcTweakMask)
+        for ss in spectra:
```
hassanxp, line 429: So each fiber is shifted independently of one another? Ok, I guess a consistent update to a detectormap both spatially and spectrally is the topic of future work.

python/pfs/drp/stella/reduceArcTask.py
```diff
@@ -395,6 +409,75 @@ def updateCenters(self, fiberTraces, detectorMap):
...
+            Output detectorMap; modified.
+        inDetectorMapList : iterable of `pfs.drp.stella.DetectorMap`
+            List of input detectorMaps.
+        """
+        num = len(inDetectorMapList)
+        for fiberId in outDetectorMap.fiberId:
+            xCenter = inDetectorMapList[0].getXCenter(fiberId)
+            for detMap in inDetectorMapList[1:]:
+                xCenter += detMap.getXCenter(fiberId)
+            xCenter /= num
```
hassanxp, line 478: Are we guaranteed that there is at lest one detectormap present?

PR #127, opened 2020-10-05
PIPE2D-622: Measure and apply small trace offsets to fiberTrace and detectorMap

PaulPrice:


hassanxp (approved):
A number of changes, but all seem sensible to me.

PR #131, opened 2020-11-12
DAMD-89: Document and implement DetectorMap

PaulPrice:
Based on tickets/PIPE2D-641.


python/pfs/drp/stella/utils.py
```diff
@@ -188,3 +192,30 @@ def getPfsVersions(prefix="VERSION_"):
         importlib.import_module(module + ".version")
         versions[prefix + name] = sys.modules[module + ".version"].__version__
     return versions
+
+
+def headerToMetadata(header):
+    """Convert FITS header to LSST metadata
```
hassanxp, line 198: Isn't this something we should move to `pfs_utils`?

hassanxp (approved):
Only minor comments.

python/pfs/drp/stella/utils.py
PaulPrice: `pfs_utils` should know nothing about LSST.

PR #132, opened 2020-11-24
PIPE2D-655: Support all spectrograph detectors

PaulPrice:


hassanxp (approved):
Fairly trivial changes - no issue here.

PR #135, opened 2020-12-08
tickets/pipe2d-661: Fix usePixels; support plt.legend()

RobertLuptonTheGood:


hassanxp (approved):
No comments. Works fine.

PR #138, opened 2020-12-09
PIPE2D-675: detMap.findPoint(fid, wavelength) returns a valid point even if wavelength is not on chip

PaulPrice:


RobertLuptonTheGood (approved):
Looks good.  And works

PR #142, opened 2021-01-19
PIPE2D-691: Fix sky subtraction for pfsObject

PaulPrice:


python/pfs/drp/stella/mergeArms.py
```diff
@@ -94,23 +92,14 @@ def runDataRef(self, expSpecRefList):
...
         if self.config.doSubtractSky1d:
-            if sky1d is None:
-                assert not self.config.doSubtractSky1dBeforeMerge
-
-                lsf = [None for _ in range(len(merged))]  # total hack!
-                sky1d = self.subtractSky1d.estimateSkyFromMerged(merged, pfsConfig, lsf)
-                expSpecRefList[0][0].put(sky1d, "sky1d")
+            lsf = [None for _ in range(len(merged))]  # total hack!
+            sky1d = self.subtractSky1d.estimateSkyFromMerged(merged, pfsConfig, lsf)
+            expSpecRefList[0][0].put(sky1d, "sky1d")
```
hassanxp, line 102: RHL mentioned in PIPE2D-680, that as we are using a simpler algorithm for the sky subtraction, we should do the subtraction post-merge. But whenever the algorithm is, then the subtraction should revert to pre-merge. I think that's why the ``doSubtractSky1dBeforeMerge`` option was introduced. Given you have removed the option, is your intention to revert the order of subtraction when we have that improved sky subtraction algorithm?

hassanxp (approved):

python/pfs/drp/stella/mergeArms.py
PaulPrice: We can certainly revert this change when we improve the sky subtraction algorithm. In the mean time, I think we benefit from the clarity that comes from removing this stub.

RobertLuptonTheGood: I don't agree.  There are two places to do the sky subtraction, and I think we need the flexibility to switch between them.  I didn't think that the code was that convoluted, but if you want to clean it up some other way that's fine with me.

PaulPrice: Sorry, I pulled the trigger before being notified of your comment. I'm happy to revert it if you like.

RobertLuptonTheGood: I think we should.  We'll definitely be putting this block (or its moral equivalent) back when we work on sky subtraction, including the work on EMPCA that I posted to slack

PR #145, opened 2021-01-29
PIPE2D-689: Vector version of DetectorMap.findPoint fails

PaulPrice:


hassanxp (approved):
Fine with me. Mostly search/replace changes

PR #149, opened 2021-02-18
PIPE2D-693: Flux errors near the dichroic are underestimated

PaulPrice:


hassanxp (approved):
This is fine - it was a little confusing that both ``FiberTraceSet`` and ``FiberProfileSet`` co-exist and have methods with the same name - ``extractSpectra()`` but OK.

PR #162, opened 2021-03-06
PIPE2D-760: Make doExtractSpectra=False work in reduceExposure

RobertLuptonTheGood:
Also skip the (currently slow) read of unnecessary calibrations


hassanxp (approved):
Changes make logical sense.

PR #166, opened 2021-03-08
PIPE2D-767: Please add an option to reduceArc to _not_ write a detectorMap

RobertLuptonTheGood:


python/pfs/drp/stella/reduceArc.py
```diff
@@ -199,6 +206,9 @@ def gather(self, dataRefList, results):
         if len(results) == 0:
             raise RuntimeError("No input spectra")
 
+        if not self.config.doUpdateDetectorMap:
+            self.log.fatal("you can't get here")
```
hassanxp, line 210: While this should never happen, a more meaningful message would be helpful, eg., "Something's wrong: code is creating a new detectorMap although doUpdateDetectorMap=False"

hassanxp (approved):

python/pfs/drp/stella/reduceArc.py
PaulPrice: An `assert` or `raise AssertionError` might be appropriate.

PR #167, opened 2021-03-08
PIPE2D-743 Misc minor fixes to drp.stella.utils

RobertLuptonTheGood:


hassanxp (approved):
No obvious problems seen.

PR #173, opened 2021-03-11
PIPE2D-761: Restructure reduceExposure to process visit one-by-one

PaulPrice:


hassanxp (approved):
Trivial fix, and fine.

PR #178, opened 2021-03-17
PIPE2D-786: DifferentialDetectorMap produces OutOfRangeError instead of NaN

PaulPrice:


src/DetectorMap.cc
```diff
@@ -118,14 +118,30 @@ DetectorMap::Array2D DetectorMap::getXCenter() const {
 }
 
 
+lsst::geom::Point2D DetectorMap::findPoint(int fiberId, double wavelength) const {
+    double const NaN = std::numeric_limits<double>::quiet_NaN();
+    lsst::geom::Point2D point;
+    try {
+        point = findPointImpl(fiberId, wavelength);
+    } catch (...) {
```
hassanxp, line 126: You are swallowing all exceptions here. You should probably catch the individual exceptions, or at minimum log the exception caught.

src/DetectorMap.cc
```diff
@@ -141,22 +157,31 @@ ndarray::Array<double, 2, 1> DetectorMap::findPoint(
...
     }
     return out;
 }
 
 
+double DetectorMap::findWavelength(int fiberId, double row) const {
+    try {
+        return findWavelengthImpl(fiberId, row);
+    } catch (...) {
+        return std::numeric_limits<double>::quiet_NaN();
```
hassanxp, line 172: See above comment.

hassanxp (approved):
Changes are fine, but at least log all exceptions caught.

src/DetectorMap.cc
PaulPrice: Swallowing all exceptions was the whole idea, but you're right that swallowing it might make it difficult to figure out why there's a problem. I suggest we make this behaviour configurable. The default will be to quietly swallow all exceptions, but there will be a way to allow exceptions to propagate, so debugging will reveal the problem.

PR #179, opened 2021-03-17
PIPE2D-787: Invalid points can crash the DetectorMap code

PaulPrice:


hassanxp (approved):
All fine. Good point about converting assertions to exceptions.

PR #180, opened 2021-03-18
PIPE2D-782: Spectra extracted from data used to construct fiberTraces are not flat

PaulPrice:


hassanxp (approved):
Implementation is as explained in the commit, and makes sense.

PR #185, opened 2021-04-01
DAMD-101: Save the outputs of shuffle

hassanxp:
A GuideStars instance (or None) now needs to be passed to the
PfsDesign and PfsConfig constructors.

### DIFF
--- a/python/pfs/drp/stella/synthetic.py
+++ b/python/pfs/drp/stella/synthetic.py
@@ -6,7 +6,7 @@ import lsst.geom
 import lsst.afw.image
 from lsst.pex.config import Config, Field
 from lsst.pipe.base import Struct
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/synthetic.py
```diff
@@ -329,4 +329,4 @@ def makeSyntheticPfsConfig(config, pfsDesignId, visit, rng=None,
                      fiberId, tract, patch, ra, dec, catId, objId, targetType, fiberStatus,
                      fiberFlux, psfFlux, totalFlux,
                      fiberFluxErr, psfFluxErr, totalFluxErr,
-                     filterNames, pfiCenter, pfiNominal)
+                     filterNames, pfiCenter, pfiNominal, None)
```
PaulPrice, line 332: I really don't like having to pass in `None`.

PR #188, opened 2021-04-17
PIPE2D-810: Allow the user to specify a pfsConfig file to be used by mergeArms

RobertLuptonTheGood:
I decided to go with only adding the option to mergeArms

### DIFF
--- a/python/pfs/drp/stella/mergeArms.py
+++ b/python/pfs/drp/stella/mergeArms.py
@@ -1,11 +1,13 @@
 from collections import defaultdict
+import os.path
+import re
 import numpy as np
 from lsst.pex.config import Config, Field, ConfigurableField, ListField, ConfigField
 from lsst.pipe.base import CmdLineTask, ArgumentParser, TaskRunner, Struct
 
 from pfs.datamodel.masks import MaskHelper
 from pfs.datamodel.wavelengthArray import WavelengthArray
-from .datamodel import PfsMerged
+from .datamodel import PfsConfig, PfsMerged
 from pfs.datamodel import Identity
 from .subtractSky1d import SubtractSky1dTask
 from .utils import getPfsVersions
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/mergeArms.py
```diff
@@ -39,6 +41,8 @@ class MergeArmsConfig(Config):
     doBarycentricCorr = Field(dtype=bool, default=True, doc="Do barycentric correction?")
     mask = ListField(dtype=str, default=["NO_DATA", "CR", "INTRP", "SAT"],
                      doc="Mask values to reject when combining")
+    pfsConfigFile = Field(dtype=str, default="", doc="""Full pathname of pfsCalib file to use.
+    If of the form "pfsConfig-0x%x-%d.fits", the pfsDesignId and visit0 will be deduced from the filename""")
```
hassanxp, line 45: If not of that form, the default pfsDesignId and visit0 values should be mentioned in the docString (I believe 666 and 0 respectively).

python/pfs/drp/stella/mergeArms.py
```diff
@@ -151,7 +155,21 @@ def runDataRef(self, expSpecRefList):
...
+        if self.config.pfsConfigFile:
+            try:
+                pfsDesignId, visit0 = re.split(r"[-.]", os.path.split(self.config.pfsConfigFile)[1])[1:-1]
+
+                pfsDesignId = int(pfsDesignId, 16)
+                visit0 = int(visit0)
+            except ValueError:
+                pfsDesignId = 666
+                visit0 = 0
+
```
hassanxp, line 167: Is there a physical pfsDesign file available with that designId?

hassanxp (commented):
Changes look OK, but need more understanding how the default values for ``designId`` and ``visit0`` are handled.

python/pfs/drp/stella/mergeArms.py
RobertLuptonTheGood: In general not, but it doesn't actually matter.  The only reason for needing them is to make the pfsConfig I/O API happy

PR #191, opened 2021-05-03
PIPE2D-796: Allow adjustment of detectorMap in reduceExposure

PaulPrice:


hassanxp (approved):
There's quite a significant amount of refactoring done here, but looks OK to me.

PR #196, opened 2021-06-11
PIPE2D-829: Investigate/fix line intensity measurements

PaulPrice:


hassanxp (approved):
Lots of changes, but well documented and understood. So looks fine. Maybe the adjusted detectorMap could be written out for debug purposes, as per ``reduceExposure``, but that can be a separate ticket.

PR #200, opened 2021-06-17
PIPE2D-853: Fix bootstrap following rename of guessedIntensity

PaulPrice:
Was renamed to 'intensity'.


hassanxp (approved):
Trivial changes. And all OK.

PR #203, opened 2021-06-30
PIPE2D-864: bootstrap does not identify lines correctly for med-res data

PaulPrice:


python/pfs/drp/stella/readLineList.py
```diff
@@ -14,17 +14,12 @@ class ReadLineListConfig(Config):
                               default=["Ar.txt", "Hg.txt", "Kr.txt", "Ne.txt", "Xe.txt", "skyLines.txt"])
     restrictByLamps = Field(dtype=bool, default=True,
                             doc="Restrict linelist by the list of active lamps? True is appropriate for arcs")
-    lampList = ListField(dtype=str, doc="list of species in lamps", default=[])
+    lampList = ListField(dtype=str, doc="list of species in lamps; overrides the header", default=[])
```
hassanxp, line 17: So we can override the lamp list, and then subsequently filter on that lamp list by the list of lamps that we believe are active?

hassanxp (approved):
This all looks fine. Question on lamp configuration added as a comment.

python/pfs/drp/stella/readLineList.py
PaulPrice: We have a list of active lamps from the header. That list can be overridden with that configuration parameter. Whatever is in the list is used to filter the line list (if `restrictByLamps`).

PR #204, opened 2021-07-01
PIPE2D-852: Added fitBroadbandSED.py

SogoMineo:
FitBroadbandSEDTask is a task to derive a probability distribution
function from broad-band SED fitting of a standard star.
The original program was provided by Takuji Yamashita.

Requires PIPE2D-852 upgrade to drp_pfs_data.

### DIFF
--- a/python/pfs/drp/stella/fitBroadbandSED.py
+++ b/python/pfs/drp/stella/fitBroadbandSED.py
@@ -1,0 +1,167 @@
+from pfs.datamodel.pfsConfig import TargetType
+
+from lsst.pex.config import Config, Field, DictField
+from lsst.pipe.base import Task
+from lsst.utils import getPackageDir
+
+import numpy
+import numpy.lib.recfunctions
+import astropy.io.fits
+
+import os
+
+
+class FitBroadbandSEDConfig(Config):
+    """Configuration for FitBroadbandSEDTask
+    """
+
+    fluxLibraryPath = Field(
+        dtype=str,
+        default=os.path.join(getPackageDir("drp_pfs_data"), "fluxCalibration",
+                             "syntheticFlux_withInterp_ps_HSCPS1GaiaSDSS.fits"),
+        doc="Synthetic photometry table"
+    )
+
+    filterMappings = DictField(
+        keytype=str, itemtype=str,
+        default={
+            "g": "HSCg", "r": "HSCr", "r2": "HSCr2", "i": "HSCi", "i2": "HSCi2",
+            "z": "HSCz", "y": "HSCy"
+        },
+        doc="Conversion table from pfsConfig's filter names to those used by `fluxLibrary`"
+    )
+
+
+class FitBroadbandSEDTask(Task):
+    """Fit an observed SED with synthetic model SEDs
+    """
+
+    ConfigClass = FitBroadbandSEDConfig
+    _DefaultName = "fitBroadbandSED"
+
+    def runDataRef(self, dataRef):
+        """For each spectrum,
+        calculate probabilities of model SEDs matching the spectrum.
+
+        Only spectra for which ``targetType == TargetType.FLUXSTD``
+        are processed.
+
+        Parameters
+        ----------
+        dataRef : `lsst.daf.persistence.ButlerDataRef`
+            Data reference for merged spectrum.
+
+        Returns
+        -------
+        broadbandPDF : `list` of `numpy.array` of `float`
+            ``broadbandPDF[iSpectrum][iSED]`` is the normalized probability
+            of the SED ``iSED`` matching the spectrum ``iSpectrum``.
+            If the targetType of ``iSpectrum`` is not ``TargetType.FLUXSTD``,
+            ``broadbandPDF[iSpectrum]`` is None.
+        """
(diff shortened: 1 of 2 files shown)


python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
...
+        doc="Synthetic photometry table"
+    )
+
+    filterAliases = DictField(
+        keytype=str, itemtype=str,
+        default={
+            "g": "HSCg", "r": "HSCr", "r2": "HSCr2", "i": "HSCi", "i2": "HSCi2",
+            "z": "HSCz", "y": "HSCy"
+        },
+        doc="Conversion table from AFW's filter names to those used by `fluxLibrary`"
```
PaulPrice, line 29: I don't think we need to worry about AFW's filter names, and there are a lot of changes to AFW's filter handling coming soon so we should avoid it. All we need to worry about are the filter names in the pfsConfig.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
...
+        pfsConfig = dataRef.get("pfsConfig")
+
+        pdfs = []
+        for filterNames, fiberFlux, fiberFluxErr \
+                in zip(pfsConfig.filterNames, pfsConfig.fiberFlux, pfsConfig.fiberFluxErr):
+            pdfs.append(self.run(filterNames, fiberFlux, fiberFluxErr))
+
+        return pdfs
+
+    def run(self, filterNames, fiberFlux, fiberFluxErr):
```
PaulPrice, line 64: Can we have the `run` method simply take a `pfsConfig`? That makes it easier for users to call directly.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    fluxLibraryPath = Field(
+        dtype=str,
+        default=os.path.join(getPackageDir("drp_pfs_data"), "fluxCalibration",
+                             "syntheticFlux_withInterp_ps_HSCPS1GaiaSDSS.fits"),
+        doc="Synthetic photometry table"
+    )
+
+    filterAliases = DictField(
```
PaulPrice, line 23: I'd call this `filterMappings`: it's a mapping from the pfsConfig filter name to the flux library column. But not really important.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
...
+        if not (len(filterNames) == len(fiberFlux) == len(fiberFluxErr)):
+            raise ValueError("Lengths of arguments must be equal.")
+        if not filterNames:
+            nSEDs = len(self.fluxLibrary)
+            return numpy.full(shape=(nSEDs,), fill_value=1.0/nSEDs, dtype=float)
+
+        # Convert filter names.
+        filterNames = [self.config.filterAliases.get(f, f) for f in filterNames]
+
+        # fluxLibrary.shape = (nSEDs, nBands)
```
PaulPrice, line 91: This looks like code, and having commented-out code is a violation of the style guide, so I suggest just prefixing it with "Note:". Ditto below.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
+from lsst.pex.config import Config, Field, DictField
+from lsst.pipe.base import Task
+from lsst.utils import getPackageDir
+
+import numpy
```
PaulPrice, line 5: We usually `import numpy as np`, but it's up to you.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -0,0 +1,135 @@
...
+        Returns
+        -------
+        broadbandPDF : `list` of `numpy.array` of `float`
+            ``broadbandPDF[iSpectrum][iSED]`` is the normalized probability
+            of the SED ``iSED`` matching the spectrum ``iSpectrum``.
+        """
+        pfsConfig = dataRef.get("pfsConfig")
+
+        pdfs = []
+        for filterNames, fiberFlux, fiberFluxErr \
```
PaulPrice, line 58: Don't you need to select the fluxcal fibers, rather than iterating over all fibers?

PaulPrice (approved):
I suggest a few changes, but this is great!

python/pfs/drp/stella/fitBroadbandSED.py
SogoMineo: `pfsConfig` contains multiple spectra, but `run()` here deals with one spectra per call. Should I change `run()` so it will take `pfsConfig`, and rename the old `run()` to `getProbabilities()`?

PaulPrice: Yeah, that sounds good!

SogoMineo:
I have changed `run()` method so that it now deals with pfsConfig. Fibers whose `targetType` is not `TargetType.FLUXSTD` are skipped. I also changed the unit test largely.

PR #210, opened 2021-07-27
PIPE2D-839: Make AdjustDetectorMapTask faster

PaulPrice:


hassanxp (approved):
More strictly speaking this is a (flexible) ``low-order`` solver rather than a linear solver?

PaulPrice:
Fitting a polynomial (no matter the order) uses a linear solver, because the equations are linear in the coefficients. This means that you can simply solve a matrix equation. Non-linear solvers are required to fit more complicated functions, and they are usually slower because you have to iterate down the slope to the minimum.
So I think it's right to say this is a linear solver.

PR #211, opened 2021-07-28
PIPE2D-797: add _DefaultName to Tasks that lacked it

PaulPrice:


hassanxp (approved):
Trivial changes. All of which are fine.

PR #218, opened 2021-09-17
PIPE2D-871: Add EstimateRadialVelocityTask

SogoMineo:
This is supposed to be a subtask of flux calibration
being worked on by Takuji Yamashita.

### DIFF
--- a/python/pfs/drp/stella/estimateRadialVelocity.py
+++ b/python/pfs/drp/stella/estimateRadialVelocity.py
@@ -1,0 +1,171 @@
+from lsst.pex.config import Config, Field, ChoiceField, ListField
+from lsst.pipe.base import Struct, Task
+from pfs.drp.stella.datamodel.interpolate import interpolateFlux
+
+import numpy as np
+from astropy import constants as const
+import scipy.optimize
+
+import math
+
+
+class EstimateRadialVelocityConfig(Config):
+    """Configuration for EstimateRadialVelocityTask
+    """
+
+    findMethod = ChoiceField(
+        doc="Peak-finding method.",
+        dtype=str,
+        allowed={
+            "peak": "The sampled point at which the cross-correlation is maximum.",
+            "gauss": "Peak of a Gaussian fit to the cross-correlation.",
+        },
+        default="gauss",
+        optional=False
+    )
+
+    searchMin = Field(
+        doc="Minimum of searched range of radial velocity, in km/s.",
+        dtype=float,
+        default=-500
+    )
+
+    searchMax = Field(
+        doc="Maximum of searched range of radial velocity, in km/s.",
+        dtype=float,
+        default=500
+    )
+
+    searchStep = Field(
+        doc="Step of searched range of radial velocity, in km/s."
+            " The actual step may be slightly smaller than this value.",
+        dtype=float,
+        default=5.0
+    )
+
+    peakRange = Field(
+        doc="Velocity range, in km/s, used in fitting gaussian (valid when `findMethod` = \"gauss\")",
+        dtype=float,
+        default=100,
+    )
+
+    useCovar = Field(
+        doc="Whether to use covariance. If False, use variance only."
+            " Covariance used, the returned error bar will be more correct,"
+            " but this task will be far less robust.",
+        dtype=bool,
+        default=True,
+    )
+
+    mask = ListField(
+        doc="Mask planes for bad pixels",
+        dtype=str,
+        default=["BAD", "SAT", "CR", "NO_DATA"],
+    )
+
+
+class EstimateRadialVelocityTask(Task):
+    """Estimate the radial velocity.
+    """
+
+    ConfigClass = EstimateRadialVelocityConfig
+    _DefaultName = "estimateRadialVelocity"
+
+    def run(self, spectrum, modelSpectrum):
+        """Get the radial velocity of ``spectrum``
+        in comparison with ``modelSpectrum``.
+
+        Parameters
+        ----------
+        spectrum : `pfs.datamodel.pfsFiberArray.PfsFiberArray`
+            Observed spectrum.
+            It must be whitened (Continuum is 1.0 everywhere.)
+        modelSpectrum : `pfs.datamodel.pfsSimpleSpectrum.PfsSimpleSpectrum`
+            Model spectrum as ``spectrum`` would be
+            were it not for the radial velocity.
+            It must be whitened (Continuum is 1.0 everywhere.)
+
+        Returns
+        -------
+        velocity : `float`
+            Radial velocity in km/s.
+        error : `float`
+            Standard deviation of ``velocity``.
+            This is reliable only if ``config.findMethod="gauss"``
+            and ``config.useCovar=True``.
+        """
+        # TODO: This method should be wholly rewritten so that it will use
+        # a log-scaled wavelength for the sake of FFT convolution.
+        searchMin = self.config.searchMin
+        searchMax = self.config.searchMax
+        searchStep = self.config.searchStep
+        searchNum = 1 + int(math.ceil((searchMax - searchMin) / searchStep))
+        searchVelocity = np.linspace(searchMin, searchMax, num=searchNum, endpoint=True)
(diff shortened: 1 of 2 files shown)


tests/test_estimateRadialVelocity.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Returns
+        -------
+        spectrum : `pfs.datamodel.pfsFiberArray.PfsFiberArray`
+        """
+        refWavelength = refSpectrum.wavelength
+        refFlux = refSpectrum.flux
+
+        # Select any 3nm range made of an odd number of sampling points
+        index = np.nonzero((800 <= refWavelength) & (refWavelength <= 803))[0]
```
PaulPrice, line 82: Can you make this range configurable? It's usually good to avoid hard-coded "magic numbers".

tests/test_estimateRadialVelocity.py
```diff
@@ -0,0 +1,174 @@
...
+        index = index[: (len(index) + 1) // 2 * 2 - 1]
+        lsfWavelength = refWavelength[index]
+
+        lsfFlux = createLsf(lsfWavelength, lsfWavelength[len(lsfWavelength)//2])
+        lsfFlux /= np.sum(lsfFlux)
+
+        c = 299792.458
```
PaulPrice, line 89: Not using astropy constants here?

tests/test_estimateRadialVelocity.py
```diff
@@ -0,0 +1,174 @@
...
+            np.zeros(0, dtype=int),
+            np.zeros(shape=(0, 2), dtype=float),
+            np.zeros(shape=(0, 2), dtype=float)
+        )
+        mask = np.zeros(len(wavelength), dtype=int)
+        sky = np.zeros(len(wavelength), dtype=float)
+        covar = np.zeros(shape=(3, len(wavelength)), dtype=float)
+        covar[0, :] = variance
+        covar2 = np.zeros(shape=(0, 0), dtype=float)
+        flags = MaskHelper()
```
PaulPrice, line 114: You should include some masked pixels.

tests/test_estimateRadialVelocity.py
```diff
@@ -0,0 +1,174 @@
...
+        (For example, ``wavelength.shape == (1, N)``
+        and ``peak.shape == (M, 1)``,
+        in which case the return value has shape ``(M, N)``.)
+
+    Returns
+    -------
+    flux : `numpy.array`
+        An LSF whose peak is at `peak`.
+    """
+    a = 6.7
```
PaulPrice, line 140: Make this configurable.

You could use `pfs.drp.stella.GaussianLsf`?

tests/test_estimateRadialVelocity.py
```diff
@@ -0,0 +1,174 @@
...
+        (For example, ``wavelength.shape == (1, N)``
+        and ``peak.shape == (M, 1)``,
+        in which case the return value has shape ``(M, N)``.)
+
+    Returns
+    -------
+    flux : `numpy.array`
+        An LSF whose peak is at `peak`.
+    """
+    a = 66
```
PaulPrice, line 165: Make configurable.

PaulPrice (approved):
A few things to take care of, but this is in good shape!

python/pfs/drp/stella/estimateRadialVelocity.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class EstimateRadialVelocityTask(Task):
+    """Estimate the radial velocity.
+    """
+
+    ConfigClass = EstimateRadialVelocityConfig
+    _DefaultName = "estimateRadialVelocity"
+
+    def getRadialVelocity(self, spectrum, modelSpectrum):
```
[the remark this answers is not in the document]
SogoMineo: What if I rename `getRadialVelocity()` to `run()` directly?

python/pfs/drp/stella/estimateRadialVelocity.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Parameters
+    ----------
+    covar : `numpy.array`
+        Covariance matrix.
+        Shape [K, N], N being the number of random variables.
+        `covar[0, :]` is the diagonal part of the matrix,
+        `covar[1, :-1]` is +-1 off diagonal,
+        `covar[2, :-2]` is +-2 off diagonal,
+        and so on.
```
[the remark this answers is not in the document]
SogoMineo: I didn't know that. I will delete this function, for the problem is much simpler.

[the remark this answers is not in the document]
SogoMineo: `pfs/datamodel/utils.py` has `makeFullCovariance` function, which I think expects the covariance between neighbouring wavelength as its argument. This is why I first wrote this function this way.

python/pfs/drp/stella/estimateRadialVelocity.py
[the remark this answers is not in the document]
PaulPrice: I think that's the best option.

python/pfs/drp/stella/estimateRadialVelocity.py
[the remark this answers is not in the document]
PaulPrice: Yeah, I'm not absolutely sure what we have now is what was originally intended, but it's what we have. I don't believe `makeFullCovariance` is used anywhere, and I'm fairly sure it outdates my working on PFS.
One day, we'll have a talk about what the various covariance bits and pieces are actually intended to be and what they should be versus what they are, but I don't think it's a priority right now.

tests/test_estimateRadialVelocity.py
SogoMineo: Could you suggest the names of flags for this task to observe? I don't know whether the list of all flag names exists or not.

PaulPrice: Here's a good example: https://github.com/Subaru-PFS/drp_stella/blob/master/python/pfs/drp/stella/photometerLines.py#L44

tests/test_estimateRadialVelocity.py
SogoMineo: This function is named LSF but actually its return value is used to dig down a fake continuous spectrum to imitate an absorption line. Because of how it is implemented, it is necessary that the height of the returned peak is 1, unlike `GaussianLsf`.

tests/test_estimateRadialVelocity.py
SogoMineo: Thank you. I talked with Yamashita-san and he said flux values have been properly interpolated before this task is called, so this task does not need to see masks. I will write such in the docstrings and tell him your link.

PaulPrice: I don't know how you get observed spectra without masked pixels. I think you need to deal with them.

python/pfs/drp/stella/estimateRadialVelocity.py
```diff
@@ -0,0 +1,178 @@
...
+            fitVelocity = searchVelocity[fitIndex]
+            fitCcf = coeff * ccf[fitIndex]
+            fitCovar = transformCovar((coeff * coeff) * covar, scaledModel[fitIndex, :])
+            if not self.config.useCovar:
+                # `scipy.optimize.curve_fit()` takes standard deviation
+                # rather than variance if `sigma` argument is not a matrix.
+                fitCovar = np.sqrt(np.diag(fitCovar))
+
+            iniParam = [1.0, velocity, self.config.peakRange]
+            pfit, pcov = scipy.optimize.curve_fit(
```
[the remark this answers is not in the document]
SogoMineo: I checked it, but the variance (covariance) argument is essential here. Yamashita-san says he will use, if available, the velocity error (`sqrt(pcov[1])`) somewhere in the latter part of the analysis. (To make the velocity error correct, variance is not sufficient. Covariance matters very much here because the sample points fit to are cross-correlation)

SogoMineo:
I made changes according to the comments. I will merge this branch in a day if there are no further comments.

PaulPrice:
Go for it! Thanks.

PR #222, opened 2021-09-28
PIPE2D-906: Fix weekly breakage in reduceExposure

PaulPrice:


python/pfs/drp/stella/readLineList.py
```diff
@@ -90,6 +90,8 @@ def filterByIntensity(self, lines):
         filtered : `pfs.drp.stella.ReferenceLineSet`
             Filtered list of reference lines.
         """
+        if self.config.minIntensity <= 0:
```
hassanxp, line 93: It might be useful to mention in the docString and/or in the description for that parameter, that no filtering is done if the value of that parameter is <= 0.

hassanxp (approved):
Looks fine. Minor comment on docString added.

PR #223, opened 2021-09-29
PIPE2D-908: Fix test breakage from PIPE2D-871

PaulPrice:


hassanxp (approved):
Change looks fine to me.

SogoMineo (approved):
It looks good, and this patch actually works with the version of numpy I use. Sorry for my test breakage.

PR #224, opened 2021-10-12
PIPE2D-904: Add aperture corrections

PaulPrice:


python/pfs/drp/stella/referenceLine.py
```diff
@@ -21,6 +23,26 @@ class ReferenceLineStatus(Bitmask):
...
 
+    @classmethod
+    def getMasks(cls):
+        """Return the corresponding `MaskHelper`
+
+        for representing the `ReferenceLineStatus` as a bitmask in a
+        `pfs.datamodel.PfsFiberArraySet` or similar.
+        """
+        def findBitSet(bitmask: ReferenceLineStatus) -> int:
+            """Return the position of the only single bit set, or zero
```
hassanxp, line 34: So the assumption here is that only one single bit would be set in the ``ReferenceLineStatus``? Why is that?

hassanxp (commented):
I think all the changes look fine, but I'm confused with why only a single ReferenceLineStatus bit is assumed to be set, and not a mix. Please clarify.

python/pfs/drp/stella/referenceLine.py
hassanxp: Following the discussion on slack I finally understand what this code is doing (mapping mask names to bit positions). This is fine with me.

PR #228, opened 2021-11-06
PIPE2D-933: Fix normalisation of extracted spectra

PaulPrice:


python/pfs/drp/stella/apertureCorrections.py
```diff
@@ -309,7 +309,7 @@ def calculateApertureCorrection(apCorr: FocalPlaneFunction, fiberId: int, wavele
         ``fluxErr`` is not provided.
     """
     result = apCorr(wavelength, pfsConfig.select(fiberId=fiberId))
-    with np.errstate(invalid="ignore", divide="ignore"):
+    with np.errstate(invalid="ignore", divide="ignore", overflow="ignore"):
```
hassanxp, line 312: I think you need to change the last param to ``over="ignore"``

hassanxp (changes requested):
The changes look fine overall, but the tests currently fail. Changes look minor however. Added a comment to fix one, and you should also remove the ``TargetType`` from the imports in ``buildFiberProfiles.py``.

PR #230, opened 2021-11-09
PIPE2D-937: Extraction is producing NaNs in the calexp

PaulPrice:


hassanxp (approved):
Changes make sense to me

PR #234, opened 2021-11-11
PIPE2D-914: Add support for SpS fast-mode exposure format

PaulPrice:


hassanxp (approved):
This change looks fine to me.

PR #235, opened 2021-11-16
PIPE2D-946: Sky subtraction doesn't take fibre sensitivity into account

PaulPrice:


hassanxp (approved):
I see no issues with the proposed changes.

PR #238, opened 2021-12-08
PIPE2D-948: reduceArc: restore the use of continuum as well as arc data

PaulPrice:


hassanxp (commented):
I can't see anything obviously wrong with the proposed changes, so fine by me.

hassanxp (approved):

PR #239, opened 2021-12-09
PIPE2D-944: Processing fails when reducing b-band continuum data and modifying detectormap

PaulPrice:


nevencaplar (commented):
I apologize for this huge delay in review of these simple tickets. Both changes seem reasonable. 
But - if I understand correctly this simply avoids the error, but it will not adjust the detectorMap in any way, right?

PaulPrice:
That's right. PIPE2D-948 contains extensive changes to support adjustment of detectorMaps using continuum.

PR #242, opened 2022-01-04
PIPE2D-921: Add class FluxModelSet

SogoMineo:
FluxModelSet is a class that helps to deal with spectra
in fluxmodeldata package.

### DIFF
--- a/python/pfs/drp/stella/fluxModelSet.py
+++ b/python/pfs/drp/stella/fluxModelSet.py
@@ -1,0 +1,136 @@
+from pfs.datamodel.masks import MaskHelper
+from pfs.datamodel.pfsSimpleSpectrum import PfsSimpleSpectrum
+from pfs.datamodel.target import Target
+
+import numpy as np
+from numpy.lib import recfunctions
+import astropy.io.fits
+
+import functools
+import os
+
+
+class FluxModelSet:
+    """A set of flux model spectra.
+
+    This class does not load all models into memory but read them from files
+    every time they are requested. This behavior is due to so vast a number
+    of the models that they cannot be contained in DRAM.
+
+    Parameters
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/fluxModelSet.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Elements of this list also contain broadband photometries
+        that are integrations of the synthetic spectra
+        with various filters applied to them.
+
+        Returns
+        -------
+        parameters : `numpy.array`
+            A structured array,
+            whose columns are SED parameters and various broadband fluxes.
```
PaulPrice, line 43: List the parameters?

tests/test_fluxModelSet.py
```diff
@@ -0,0 +1,37 @@
...
+
+@unittest.skipIf(dataDir is None, "fluxmodeldata not setup")
+class FluxModelSetTestCase(lsst.utils.tests.TestCase):
+    def setUp(self):
+        self.fluxModelSet = FluxModelSet(os.path.join(dataDir))
+
+    def testGetSpectrum(self):
+        """Test getFileName()
+        """
+        parameters = self.fluxModelSet.parameters
```
PaulPrice, line 24: Check that this contains the expected list of parameters?

PaulPrice (approved):
Awesome!

Made some suggestions for you to adopt or ignore as appropriate.

tests/test_fluxModelSet.py
SogoMineo: If any of the expected parameters ("teff", "logg", "m", "alpha") do not exist, an exception will be raised from below, which makes the test fail. I think this is already sufficient.

PR #243, opened 2022-01-05
PIPE2D-966: Investigate failed weekly 2022-01-02

PaulPrice:


hassanxp (changes requested):
I think all the changes look OK, but concerned about why the following test fails: 
   
```
def testDuplicatesExact(self):
        self.contents = [
            Line(123.456, 1000, "NeI", 0),
            Line(123.456, 1000, "KrI", 0),
            Line(567.890, 999, "ArI", 0),
        ]
        self.config.exclusionRadius = 10.0
        lineList = self.makeLineList()

        # Check that duplicate lines are marked for rejection
        self.assertEqual(lineList[0].status, ReferenceLineStatus.BLEND)
        self.assertEqual(lineList[1].status, ReferenceLineStatus.BLEND)
````

But changing the wavelength for the second line to `123.457` (ie., a small difference compared to the wavelength of the first line) results in the test passing.

PR #244, opened 2022-01-22
PIPE2D-967: showDetectorMap - fix call to pfsConfig.selectFiber()

hassanxp:
pfsConfig.selectFiber() returns a scalar int value if the input argument
is also a scalar. This is how it is used in showDetectorMap.


PaulPrice (approved):
Looks good.

PR #245, opened 2022-01-24
PIPE2D-968: ReduceExposure 2D sky subtraction fails for SuNSS exposures

PaulPrice:


hassanxp (approved):
Trivial fix works fine. Suggest adding a check to see if there are zero sky fibers available (that can be a separate ticket if that's cleaner).

PR #248, opened 2022-02-17
PIPE2D-979: 2d sky subtraction creates patches of negative data

PaulPrice:


hassanxp (approved):
All changes are fine.

PR #250, opened 2022-02-24
PIPE2D-989: Disable auto sky fiber selection by default

PaulPrice:


hassanxp (approved):
This looks fine to me.

PR #252, opened 2022-03-01
PIPE2D-988: Add flags to identify which lines were used in detectorMap and sky subtraction

PaulPrice:


python/pfs/drp/stella/apertureCorrections.py
```diff
@@ -41,7 +41,7 @@ class MeasureApertureCorrectionsConfig(Config):
     def setDefaults(self):
         super().setDefaults()
         self.apertureFlux.radii = [3.0]
-        self.fit.mask = ReferenceLineStatus.getMasks().flags.keys()
+        self.fit.mask = ["NOT_VISIBLE", "BLEND", "SUSPECT", "REJECTED", "BROAD"]
```
hassanxp, line 44: Shouldn't the ReferenceLineStatus bit mask names be used here rather than literal strings?

hassanxp (approved):
Looks fine. Minor comment on use of literal strings rather than available constants.

python/pfs/drp/stella/apertureCorrections.py
PaulPrice: No, that's [defined](https://github.com/Subaru-PFS/drp_stella/blob/3a2c02d41eea00b16afb3649f0ab3fd04e5df1bd/python/pfs/drp/stella/fitFocalPlane.py#L22-L23) as a list of strings that get converted to a bit mask [here](https://github.com/Subaru-PFS/drp_stella/blob/3a2c02d41eea00b16afb3649f0ab3fd04e5df1bd/python/pfs/drp/stella/fitFocalPlane.py#L90).

PR #253, opened 2022-03-04
PIPE2D-922: Add fitPfsFluxReference.py

SogoMineo:
fitPfsFluxReference.py is intended to be a subtask of flux calibration.
It does the task previously under the charge of fitReference.py.
This subtask runs, but is not in its final state.
To be specific, more lines of code in relation to galactic extinction must be added.

This branch requires obs_pfs/tickets/PIPE2D-922.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -1,0 +1,989 @@
+from lsst.pex.config import Config, ConfigurableField, Field, ListField
+from lsst.pipe.base import Struct, Task
+from lsst.utils import getPackageDir
+from pfs.datamodel.identity import Identity
+from pfs.datamodel.masks import MaskHelper
+from pfs.datamodel.observations import Observations
+from pfs.datamodel.pfsConfig import FiberStatus, TargetType
+from pfs.datamodel.pfsFiberArray import PfsFiberArray
+from pfs.datamodel.pfsFiberArraySet import PfsFiberArraySet
+from pfs.datamodel.pfsFluxReference import PfsFluxReference
+from pfs.datamodel.pfsSimpleSpectrum import PfsSimpleSpectrum
+from pfs.datamodel.target import Target
+from pfs.drp.stella.fluxModelInterpolator import FluxModelInterpolator
+from pfs.drp.stella import ReferenceLine, ReferenceLineSet, ReferenceLineStatus
+from pfs.drp.stella import SpectrumSet
+from pfs.drp.stella.estimateRadialVelocity import EstimateRadialVelocityTask
+from pfs.drp.stella.fitBroadbandSED import FitBroadbandSEDTask
+from pfs.drp.stella.fitContinuum import FitContinuumTask
+from pfs.drp.stella.fitReference import FilterCurve
+from pfs.drp.stella.fluxModelSet import FluxModelSet
+from pfs.drp.stella.interpolate import interpolateFlux
+from pfs.drp.stella.lsf import GaussianKernel1D
+from pfs.drp.stella.utils.psf import fwhmToSigma
+
+from astropy import constants as const
+import numpy as np
+import scipy.integrate
+
+import copy
+
+__all__ = ["FitPfsFluxReferenceConfig", "FitPfsFluxReferenceTask"]
+
+
+class FitPfsFluxReferenceConfig(Config):
+    """Configuration for FitPfsFluxReferenceTask
+    """
+
+    fitBroadbandSED = ConfigurableField(target=FitBroadbandSEDTask,
+                                        doc="Get probabilities of SEDs from broadband photometries.")
+    fitObsContinuum = ConfigurableField(target=FitContinuumTask,
+                                        doc="Fit a model to observed spectrum's continuum")
+    fitModelContinuum = ConfigurableField(target=FitContinuumTask,
+                                          doc="Fit a model to model spectrum's continuum")
+    estimateRadialVelocity = ConfigurableField(target=EstimateRadialVelocityTask,
+                                               doc="Estimate radial velocity.")
+    minWavelength = Field(
+        dtype=float,
+        default=600.0,
+        doc="min of the wavelength range in which observation spectra are compared to models.",
+    )
+    maxWavelength = Field(
+        dtype=float,
+        default=1200.0,
+        doc="max of the wavelength range in which observation spectra are compared to models.",
+    )
+    ignoredRangesLeft = ListField(
+        dtype=float,
+        default=[685.0, 716.0, 759.0, 810.0, 895.0, 1100.0],
+        doc="Left ends of wavelength ranges ignored (because e.g. of strong atomospheric absorption)"
+            " when comparing observation spectra to models."
+    )
+    ignoredRangesRight = ListField(
+        dtype=float,
+        default=[695.0, 735.0, 770.0, 835.0, 985.0, 1200.0],
+        doc="Right ends of wavelength ranges ignored (because e.g. of strong atomospheric absorption)"
+            " when comparing observation spectra to models."
+    )
+    badMask = ListField(
+        dtype=str,
+        default=["BAD", "SAT", "CR", "NO_DATA"],
+        doc="Mask planes for bad pixels"
+    )
+    modelSNR = Field(
+        dtype=float,
+        default=400,
+        doc="Supposed S/N of model spectra."
+            " Used in making up the variance of the flux for algorithms that require it."
+            " It is not that the model spectra are affected by this amount of noise,"
+            " nor is it that any artificial noise will be added to the model spectra."
+    )
+    Rv = Field(
+        dtype=float,
+        default=3.1,
+        doc="Ratio of total to selective extinction at V, Rv = A(V)/E(B-V)."
+    )
+
+    def setDefaults(self):
+        super().setDefaults()
+        # Not sure these paramaters are good.
+        self.fitObsContinuum.numKnots = 50
+        self.fitObsContinuum.doMaskLines = True
+        self.fitObsContinuum.maskLineRadius = 25
+        self.fitModelContinuum.numKnots = 50
+        self.fitModelContinuum.doMaskLines = True
+        self.fitModelContinuum.maskLineRadius = 25
+
+
+class FitPfsFluxReferenceTask(Task):
+    """Construct reference for flux calibration.
+    """
+
+    ConfigClass = FitPfsFluxReferenceConfig
+    _DefaultName = "FitPfsFluxReference"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.makeSubtask("fitBroadbandSED")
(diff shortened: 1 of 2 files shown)


SogoMineo:
I changed fitReference.py so that it requires the unit of wavelength in the response function files to be nm.

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+        Returns
+        -------
+        spectra : `pfs.datamodel.pfsFiberArraySet.PfsFiberArraySet`
+            The same instance as the argument.
+        """
+        # Mask atmospheric absorption lines etc.
+        wavelength = spectra.wavelength
+        badMask = spectra.flags.add("ATMOSPHERE")
+
+        for m, M in zip(self.config.ignoredRangesLeft, self.config.ignoredRangesRight):
```
PaulPrice, line 655: `low, high`?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+            if bestParam is None:
+                bestParam = tuple(peakParam[name] for name in paramNames + fixedParamNames)
+
+            wavelength, flux = self.modelInterpolator.interpolate(*bestParam)
+
+            target = Target(0, 0, "0,0", 0)
+            mask = np.zeros(shape=wavelength.shape, dtype=int)
+            flags = MaskHelper()
+            mask[:] = np.where(np.isfinite(flux), 0, flags.add("BAD"))
+            spectrum = PfsSimpleSpectrum(target, wavelength, flux, mask, flags)
```
PaulPrice, line 633: I thought this was supposed to be a `PfsFluxReference` or something like that? I'm concerned that the need to lie to the datamodel (a fake `Target`) indicates that there's something missing from the datamodel.

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+        }
+
+        # Whiten spectra
+        for index, fiberId in enumerate(spectra.fiberId):
+            continuum = continua.get(fiberId)
+            if continuum is not None:
+                spectra.flux[index, :] /= continuum.flux
+                spectra.covar[index, 0, :] /= np.square(continuum.flux)
+            else:
+                spectra.mask[index, :] &= spectra.flags.add("BAD")
```
PaulPrice, line 347: Are you sure you don't want `|=`?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+
+    Returns
+    -------
+    spectrum : `pfs.datamodel.PfsSimpleSpectrum`
+        The same instance as the argument.
+    """
+    fwhm = 0.2  # typical FWHM, in nm, of LSF.
+    n = len(spectrum.wavelength)
+    dlambda = spectrum.wavelength[n//2 + 1] - spectrum.wavelength[n//2]
+    sigma = 0.42466090014400953 * fwhm  # 0.424... = 1/sqrt(log(256))
```
PaulPrice, line 774: Use `pfs.drp.stella.utils.psf.fwhmToSigma`.

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+    numer = np.sum((flux * modelFlux) * invVar)
+    denom = np.sum(np.square(modelFlux) * invVar)
+    ampli = numer / denom
+    chisq = np.sum(np.square(flux - ampli * modelFlux) * invVar)
+
+    return chisq
+
+
+def promoteSimpleSpectrumToFiberArray(spectrum, snr):
+    """Promote an instance of PfsSimpleSpectrum to PfsFiberArray.
```
PaulPrice, line 866: Why would we want to do this? Does this imply something lacking in the datamodel implementation?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+    fiberArraySet : `pfs.datamodel.pfsFiberArraySet.PfsFiberArraySet`
+        Set of spectra observed with a set of fibers
+
+    Yields
+    -------
+    fiberArray : `pfs.datamodel.pfsFiberArray.PfsFiberArray`
+        spectrum observed with a fiber.
+    """
+    for fiberId in pfsConfig.fiberId:
+        yield fiberArraySet.extractFiber(PfsFiberArray, pfsConfig, fiberId)
```
PaulPrice, line 958: I think you could do something like:

```
for ii, fiberId in pfsConfig.fiberId:
    yield fiberArraySet.select(fiberId=fiberId)
```

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+
+
+def fiberConfigs(pfsConfig):
+    """Iterator that yields single-fiber PfsConfigs.
```
PaulPrice, line 962: I think there's a note in `PfsConfig` saying that "we're not going to provide an iterator because it would be inefficient". Maybe it's not so bad, and this should live there?

ups/drp_stella.table
```diff
@@ -3,6 +3,7 @@ setupRequired(pipe_drivers)
 setupRequired(afw)
 setupRequired(meas_base)
 setupRequired(obs_pfs)
+setupRequired(drp_fstar)
```
PaulPrice, line 6: We need to update our scripts to handle the extra package (currently, the integration test fails). Do we really need as an extra package? As far as I can see, it adds just a few extra python scripts that could just as well live in drp_stella.

PaulPrice (commented):
We need to make sure that the integration test passes. At the moment, it fails on the attempt to `setup` drp_fstar. I don't see the value in keeping drp_fstar as a separate package; I think we should just put the contents into drp_stella. But perhaps I'm missing something?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+        if mode == "model":
+            fitContinuum = self.fitModelContinuum
+
+        # If `spectra` is actually a single spectrum,
+        # we put it into PfsFiberArraySet
+        if len(spectra.flux.shape) == 1:
+            original_spectrum = spectra
+            if not hasattr(spectra, "covar"):
+                spectra = promoteSimpleSpectrumToFiberArray(spectra, snr=400)
+            spectra = promoteFiberArrayToFiberArraySet(spectra, fiberId=1)
```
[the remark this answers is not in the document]
SogoMineo: The result object of these promotions is used only temporarily. `fiberId` can be any number. I chose "1" because I was afraid that `fiberId=0` might possibly have some special meaning.

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -0,0 +1,978 @@
...
+
+        # This function actually works with `PfsFiberArraySet`
+        # nonetheless for its name.
+        # (PfsArm is a subclass of PfsFiberArraySet)
+        specset = SpectrumSet.fromPfsArm(spectra)
+
+        lines = ReferenceLineSet.fromRows([
+            ReferenceLine("Hbeta", 486.1, 1.0, ReferenceLineStatus.GOOD),
+            ReferenceLine("Halpha", 656.3, 1.0, ReferenceLineStatus.GOOD),
+        ])
```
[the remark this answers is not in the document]
SogoMineo: The values are inaccurage because they are in air. I am fixing them. There is no particular reason for the impreciseness. I will write more digits.

> should this list be more complete
I have not looked into it but Yamashita-san says that he wants to mask these two lines because they are so wide that they do not vanish with the median filter. He does not care about any other lines because they disappear with the median filter. It was I who decided to write the two line directly in the source code. I thought that a file with only two short lines would not deserve to exist separately.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I want to do this because I want to pass a model spectrum to `fitContinuum`. The model spectrum is `PfsSimpleSpectrum` because it does not have `covar`. On the other hand, `fitContinuum` expects `SpectrumSet` and `SpectrumSet.fromPfsArm` expects `PfsFiberArraySet`. For this reason, I have to promote `PfsSimpleSpectrum` to `PfsFiberArraySet`. I don't know whether there are other use cases for these type promotions. If there are, it is more natural that `datamodel` should contain the type promotion methods.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I agree that this iteration is inefficient, but its inefficiency is negligible in `FitPfsFluxReferenceTask`. The iterators (`fiberConfigs()` and `fibers()`) are convenient indeed.

SogoMineo: I don't have a strong opinion whether `PfsConfig` itself should be iterable nonetheless for the inefficiency.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: `extractFiber()` returns `PfsFiberArray`. `select()` returns `PfsFiberArraySet`. I cannot change them.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I'm trying to replace `List[PfsSimpleSpectrum]` with `PfsFluxReference`. To be frank, I think there should be a simpler `PfsSimpleSpectrum` and `PfsFluxArray(Set)` that do not have target or identity, and the current `PfsSimpleSpectrum` and `PfsFluxArray(Set)` should be subclasses of the simpler ones. There should also be a simple (no target, no identity) `PfsSimpleSpectrumSet`, of which `PfsFluxReference` should be a subclass.

ups/drp_stella.table
SogoMineo: Yamashita-san agreed that `drp_fstar` can be merged to `drp_stella`. ~Should I make another ticket~ Yamashita-san is making another ticket to merge `drp_fstar` to `drp_stella`. I will make a PR for it before working on this ticket further.

python/pfs/drp/stella/fitPfsFluxReference.py
PaulPrice: That makes sense. I agree.

SogoMineo:
I dropped the dependence on drp_fstar. I believe the changes will no longer break the integration test. As for "run() is too long" problem, I failed to divide it very cleanly, but its length is now 100 lines if the sub-function is excluded. Because the JIRA issue state has already been turned review complete, I am merging this PR.

PR #254, opened 2022-03-07
PIPE2D-993: Use only common fibers from pfsConfig, FiberTraceSet and apCorr

hassanxp:
Otherwise processing may fail if there is a mis-match.

### DIFF
--- a/python/pfs/drp/stella/subtractSky2d.py
+++ b/python/pfs/drp/stella/subtractSky2d.py
@@ -253,10 +253,25 @@ class SubtractSky2dTask(Task):
         if psf is None:
             raise RuntimeError("Unable to construct sky image: PSF is None")
         result = MaskedImageF(bbox)
-        fiberId = np.array([ft.fiberId for ft in fiberTraces])
-        centers = pfsConfig.extractCenters(fiberId)
+
+        # Compute essentially the intersection of fiberIds and traces from the 3 input sources:
+        # 1. the pfsConfig
+        # 2. the aperture correction object
+        # 3. the input fiberTraceSet.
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/subtractSky2d.py
```diff
@@ -253,10 +253,25 @@ def makeSkyImage(self, bbox, psf, fiberTraces, pfsConfig, sky2d, apCorr):
...
+        # 3. the input fiberTraceSet.
+        fiberIdsFromConfigApCorr = set(pfsConfig.fiberId).intersection(apCorr.fiberId)
+        fiberTracesMatched = [ft for ft in fiberTraces if ft.fiberId in fiberIdsFromConfigApCorr]
+        fiberIdMatched = np.array([ft.fiberId for ft in fiberTracesMatched])
+
+        excludedFiberIds = list(set(pfsConfig.fiberId) - set(fiberIdMatched))
+        self.log.infof('pfsConfig fiberIds {}, with corresponding fiberStatus {} '
+                       'are excluded from sky model image generation.',
+                       excludedFiberIds,
+                       pfsConfig.fiberStatus[np.isin(pfsConfig.fiberId, excludedFiberIds)].tolist())
```
PaulPrice, line 269: Since the pfsConfig contains information about fibers from multiple spectrographs, this log message is going to be dominated by numbers that we don't care about. It's only helpful if there are fibers that we care about that aren't being subtracted.

PaulPrice (approved):

python/pfs/drp/stella/subtractSky2d.py
hassanxp: Agreed. Will demote to ``debugf``.

PR #255, opened 2022-03-15
PIPE2D-999: 2D sky subtraction results in over subtracted images

PaulPrice:


hassanxp (approved):
Both commits are fine, thanks.

src/photometry.cc
```diff
@@ -160,7 +160,7 @@ lsst::afw::table::BaseCatalog photometer(
                 // Bad PSF: not even a blend.
                 continue;
             }
-            box.shift(lsst::geom::Extent2D(point) - 0.5*box.getDimensions());  // put the center on the point
+            box.shift(lsst::geom::Extent2D(point));  // put the center on the point
```
hassanxp, line 161: In the commit comment, you mentioned that the shifting is 'unnecessary', but it is more than unnecessary, surely, it is erroneous?

hassanxp (approved):
Agree with all the changes. Only one very minor question on one of the commit comments.

PR #258, opened 2022-03-28
PIPE2D-1010: Update spectrumset to support differing fiberIds in pfsConfig/fiberProfile

hassanxp:


PaulPrice (approved):
Trivial style fix proposed.

PR #260, opened 2022-03-29
PIPE2D-1014: Store radial velocities in PfsFluxReference objects

SogoMineo:
Radial velocities are required by the subsequent task,
and they have to be stored in PfsFluxReference objects.


PaulPrice (approved):
It would be good if there were some documentation on the units of these quantities, but maybe that doesn't belong here.

SogoMineo:
I think I can use `astropy.io.fits.FITS_rec` instead of structured array to store the units. I will try it later.

PR #261, opened 2022-03-30
PIPE2D-1016: pfsArm.flux and pfsArm.sky values differ in scale

PaulPrice:


hassanxp (approved):
Changes make sense to me.

PR #262, opened 2022-04-12
PIPE2D-1021: Make FitPfsFluxReferenceTask executable

SogoMineo:
FitPfsFluxReferenceTask is now executable, and its output is written to a file.
This branch requires `tickets/PIPE2D-1021` branch of `obs_pfs`.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -1,5 +1,5 @@
 from lsst.pex.config import Config, ConfigurableField, Field, ListField
-from lsst.pipe.base import Struct, Task
+from lsst.pipe.base import ArgumentParser, CmdLineTask, Struct
 from lsst.utils import getPackageDir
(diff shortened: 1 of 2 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -133,9 +133,10 @@ def runDataRef(self, dataRef):
         dataRef : `lsst.daf.persistence.ButlerDataRef`
             Data reference for exposure.
         """
+        self.log.info(f"Processing {dataRef.dataId}")
```
PaulPrice, line 136: For log messages, please use the `%` form of formatting, but only provide the formatting arguments rather than doing the formatting yourself. E.g.,

    self.log.info("Processing %s", dataRef.dataId)

This allows the formatting to be skipped if the log message wouldn't be printed. It's not a big deal here, since you're not in a tight loop, but it can matter in some contexts.

PaulPrice (approved):
Minor comment only.

PR #263, opened 2022-04-15
PIPE2D-1013: Add fitFluxCal.py

SogoMineo:
This executable makes "fluxCal" from "pfsFluxReference",
just like fluxCalibrate.py makes "fluxCal" from "pfsReference".

### DIFF
--- a/bin.src/fitFluxCal.py
+++ b/bin.src/fitFluxCal.py
@@ -1,0 +1,3 @@
+#!/usr/bin/env python
+from pfs.drp.stella.fitFluxCal import FitFluxCalTask
+FitFluxCalTask.parseAndRun()
--- a/python/pfs/drp/stella/fitFluxCal.py
+++ b/python/pfs/drp/stella/fitFluxCal.py
@@ -1,0 +1,174 @@
+from collections import defaultdict
+
+from astropy import constants as const
+import numpy as np
+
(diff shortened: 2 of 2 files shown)


python/pfs/drp/stella/datamodel/pfsFluxReference.py
```diff
@@ -0,0 +1,9 @@
+import pfs.datamodel.pfsFluxReference
+
+__all__ = ("PfsFluxReference",)
+
+
+class PfsFluxReference(pfs.datamodel.pfsFluxReference.PfsFluxReference):
+    """datamodels's ``PfsFluxReference`` with additional utility methods.
+    """
+    # No additional methods at present.
```
PaulPrice, line 9: If you don't need to extend the class, then you don't need this file. The reason we do it with other classes is because there are features we need for the pipeline that we don't make available for the general-purpose no-dependency datamodel package.

PaulPrice (approved):

python/pfs/drp/stella/datamodel/pfsFluxReference.py
SogoMineo: I am deleting this file, and revising `PfsMapper.yaml` in `obs_pfs`, which has been pointing to `pfs.` *drp.stella* `.PfsFluxReference` class.

PR #264, opened 2022-04-21
PIPE2D-1026: fitPfsFluxReference.py: Take `PfsMerged.norm` into account

SogoMineo:
The essential change is lines 362-366. Most of the other changes are required because I use the division operator of the extended PfsMerged class. I take this opportunity and use the division operator also at lines 403-405, though this change might be irrelevant here.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -367,6 +367,11 @@ class FitPfsFluxReferenceTask(CmdLineTask):
         if mode == "model":
             fitContinuum = self.fitModelContinuum
 
+        if hasattr(spectra, "norm"):
+            # We want the flux normalized.
+            spectra /= spectra.norm
+            spectra.norm[...] = 1.0
+
         # If `spectra` is actually a single spectrum,
         # we put it into PfsFiberArraySet
         if len(spectra.flux.shape) == 1:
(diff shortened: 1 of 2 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -149,7 +149,7 @@ def run(self, pfsConfig, pfsMerged):
         pfsConfig : `pfs.datamodel.pfsConfig.PfsConfig`
             Configuration of the PFS top-end,
             in which information of broad band fluxes count.
-        pfsMerged : `pfs.datamodel.pfsFiberArraySet.PfsFiberArraySet`
+        pfsMerged : `pfs.drp.stella.PfsFiberArraySet`
```
PaulPrice, line 152: Isn't it `pfs.drp.stella.datamodel.PfsFiberArraySet`?
I think this change should be a separate commit, as it's not relevant to the normalisation change.

Similarly throughout.

PaulPrice (approved):

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I am splitting the commit into two. In the first commit `pfs.datamodel.PfsFiberArraySet` will be replaced by `pfs.drp.stella.datamodel.PfsFiberArraySet`, and in the second commit the normalisation will be done.

PR #268, opened 2022-05-03
PIPE2D-1015: Mask bit flags not propagated to pfsArm correctly

PaulPrice:


hassanxp (approved):
Changes look fine to me.

PR #272, opened 2022-06-07
PIPE2D-1049: Deadlock in weekly processing

PaulPrice:


hassanxp (approved):
All changes look fine to me, and in any case temporary.

PR #279, opened 2022-07-12
PIPE2D-1055: Fix weekly failure 2022-06-12

PaulPrice:


hassanxp (approved):
Changes look fine to me. To me, the last two commits (computing the median while respecting masks) should be merged, but you probably had good reason to keep those commits separate so I'll leave that to you.

PR #280, opened 2022-07-27
PIPE2D-1017: added transition and source to line list

hassanxp:


python/pfs/drp/stella/referenceLine.py
```diff
@@ -63,11 +74,17 @@ class ReferenceLine:
         Estimated intensity (arbitrary units).
     status : `int`
         Bitmask indicating the quality of the line.
+    transition : `str`
+        Line transition information (or "NOT_AVAILABLE")
```
PaulPrice, line 78: I think `UNKNOWN` is what you've chosen to use when it's missing.

python/pfs/drp/stella/referenceLine.py
```diff
@@ -123,7 +140,13 @@ def fromLineList(cls, filename):
 
                 fields = line.split()
                 try:
-                    wavelength, intensity, description, status = fields
+                    # Support older linelists that do not have transition or source information
+                    if len(fields) == 4:
+                        wavelength, intensity, description, status = fields
+                        transition = None
```
PaulPrice, line 146: Shouldn't this be `"UNKNOWN"`?

PaulPrice (approved):
This is going to conflict with DAMD-97.

PR #283, opened 2022-08-08
PIPE2D-1065: fitPfsFluxReference.py: LSF convolution before whitening

SogoMineo:
LSF convolution must at all costs be done before whitening because the convolution lowers the continuum due to dense absorption lines. Because the density of the absorption lines varies from place to place, the result of LSF convolution after whitening is not even horizontal. These commits makes the two operations take place in the correct order.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -398,8 +398,8 @@ class FitPfsFluxReferenceTask(CmdLineTask):
         specset = SpectrumSet.fromPfsArm(spectra)
 
         lines = ReferenceLineSet.fromRows([
-            ReferenceLine("Hbeta", 486.2721, 1.0, ReferenceLineStatus.GOOD),
-            ReferenceLine("Halpha", 656.4614, 1.0, ReferenceLineStatus.GOOD),
+            ReferenceLine("Hbeta", 486.2721, 1.0, ReferenceLineStatus.GOOD, "", 0),
+            ReferenceLine("Halpha", 656.4614, 1.0, ReferenceLineStatus.GOOD, "", 0),
         ])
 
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -952,7 +1008,10 @@ def calculateSpecChiSquare(obsSpectrum, model, radialVelocity, badMask):
     bad = (0 != (obsSpectrum.mask & obsSpectrum.flags.get(*(m for m in badMask if m in obsSpectrum.flags))))
 
     flux = np.copy(obsSpectrum.flux)
-    flux[bad] = 0.0
+    # If not(any(isnan(flux))), we don't have to do this
+    # (because invVar will be set to 0 for bad pixels.)
+    # We set flux[bad] to a finite value to make sure we won't encounter nan.
+    flux[bad] = 1.0
```
hassanxp, line 1014: Are the 'bad' flux values ever used in processing? If so, you shouldn't encounter ``nan`` if only 'good' flux pixels are used.

hassanxp (approved):
Changes look fine. Minor comment on setting of flux values in the case of 'bad' pixels.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: Instead of excluding the bad pixels `flux[bad]`, I set the weights of these pixels to zero. These pixels are actually used (with zero weights). So, these pixels may be any value, but they must be finite. I added the three lines of comment because it took a while to recall what my intention had been here. I changed the r.h.s from 0 to 1 because the unmarked value for the whitened flux is not 0 but 1. When I plot `flux` here, `flux` looks very strange if `flux[bad]` is set to zero.

PR #291, opened 2022-10-19
PIPE2D-1098: Bootstrap a detectorMap for n2

PaulPrice:


python/pfs/drp/stella/buildFiberProfiles.py
```diff
@@ -268,9 +268,9 @@ def associatePeaks(self, peaks, width):
                 del indices[-1]
                 unassociated[pp.peak] = pp
                 for index in indices:
-                    # Score is distance to most recent trace peak + negative number of past associations
+                    # Score is negative number of past associations and distance to most recent trace peak
```
hassanxp, line 271: This comment is a little confusing as the ``score`` variable itself is till the distance to the most recent trace peak. Suggest renaming the variable.

hassanxp (approved):
Looks good. Would suggest renaming ``score`` variable and/or updating comment about new score strategy.

PR #293, opened 2022-11-02
PIPE2D-1106: Don't make PfsSingle for UNASSIGNED or ENGINEERING targets

SogoMineo:
Fibers of these targetTypes have (ra, dec) = (nan, nan) in PfsConfig. This is appropriate because these fibers do not point to the sky. But the coordinates being NaN causes an exception during the process of making PfsSingle, for NaN is not allowed in FITS headers.

### DIFF
--- a/python/pfs/drp/stella/fitFluxCal.py
+++ b/python/pfs/drp/stella/fitFluxCal.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+import math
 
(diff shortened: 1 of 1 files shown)


PaulPrice (approved):
We need to drop only `TargetType.ENGINEERING`.

PaulPrice (approved):

PR #299, opened 2022-12-01
PIPE2D-1122: Investigate poor detectorMap adjustment in r1

PaulPrice:


hassanxp (approved):
I think all the changes are fine.

hassanxp (approved):
Changes look fine, but the commit comment about flagging is a little confusing: You mention that "We use the new getExclusionZone function to FLAG blended lines as bad." but later you mention that "[...]lines AREN'T FLAGGED, which means they're not
automatically tossed downstream." Maybe if remove the "flagged" part of the second sentence to read "lines are not automatically tossed downstream" all is clear.

PR #300, opened 2022-12-02
PIPE2D-1110: Change filter names

SogoMineo:
The integration test and the weekly test will be broken after this change. To fix this problem, run
`PIPE2D-1110-change-filter-names.py`.
There is a link to this file in PIPE2D-1110, JIRA.

### DIFF
--- a/python/pfs/drp/stella/fitBroadbandSED.py
+++ b/python/pfs/drp/stella/fitBroadbandSED.py
@@ -17,8 +17,26 @@ class FitBroadbandSEDConfig(Config):
     filterMappings = DictField(
         keytype=str, itemtype=str,
         default={
-            "g": "HSCg", "r": "HSCr", "r2": "HSCr2", "i": "HSCi", "i2": "HSCi2",
-            "z": "HSCz", "y": "HSCy"
+            "g_hsc": "HSCg",
+            "r_old_hsc": "HSCr",
+            "r2_hsc": "HSCr2",
+            "i_old_hsc": "HSCi",
+            "i2_hsc": "HSCi2",
+            "z_hsc": "HSCz",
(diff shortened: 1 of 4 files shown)


python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -17,8 +17,26 @@ class FitBroadbandSEDConfig(Config):
...
+            "z_ps1": "PS1z",
+            "y_ps1": "PS1y",
+            "bp_gaia": "GaiaBp",
+            "rp_gaia": "GaiaRp",
+            "g_gaia": "GaiaG",
+            "u_sdss": "SDSSu",
+            "g_sdss": "SDSSg",
+            "r_sdss": "SDSSr",
+            "i_sdss": "SDSSi",
+            "z_sdss": "SDSSz",
```
PaulPrice, line 39: Why do we have a two-step mapping from pfsConfig to internal name to filename, rather than a one-step mapping?

PaulPrice (approved):
Could you please add an explanation of why this change is necessary in the commit message?
We need to coordinate merging with updating the integration test and weekly.

python/pfs/drp/stella/fitBroadbandSED.py
SogoMineo: The package (`fluxmodeldata-ambre...`) used by `fitBroadbandSED.py` is not version-controlled. I don't want to variegate it unnecessarily. I have an undecided plan to change `fluxmodeldata-ambre...` in near future (to add theoretically whitened flux), and if it happens, I will change the filter names used in `fluxmodeldata-ambre...` so that I can delete this mapping.

PR #303, opened 2022-12-13
PIPE2D-1127: fitPfsFluxReference: Make configurable which b.b.flux to use

SogoMineo:
`fitPfsFluxReference` has assumed that broadband fluxes are stored in `pfsConfig.fiberFlux`. In reality, they are in `pfsConfig.psfFlux`, and `pfsConfig.fiberFlux` is `NaN`. In this branch, it is a config parameter which flux to choose. `psfFlux` is chosen by default.

### DIFF
--- a/python/pfs/drp/stella/fitBroadbandSED.py
+++ b/python/pfs/drp/stella/fitBroadbandSED.py
@@ -20,8 +20,7 @@ except ImportError:
 
 
 class FitBroadbandSEDConfig(Config):
-    """Configuration for FitBroadbandSEDTask
-    """
+    """Configuration for FitBroadbandSEDTask"""
 
     broadbandFluxType = ChoiceField(
         doc="Type of broadband fluxes to use.",
@@ -32,11 +31,12 @@ class FitBroadbandSEDConfig(Config):
             "total": "Use `psfConfig.totalFlux`.",
         },
         default="psf",
-        optional=False
+        optional=False,
     )
 
     filterMappings = DictField(
-        keytype=str, itemtype=str,
+        keytype=str,
+        itemtype=str,
         default={
             "g_hsc": "HSCg",
             "r_old_hsc": "HSCr",
@@ -59,13 +59,12 @@ class FitBroadbandSEDConfig(Config):
             "i_sdss": "SDSSi",
             "z_sdss": "SDSSz",
         },
-        doc="Conversion table from pfsConfig's filter names to those used by `fluxLibrary`"
+        doc="Conversion table from pfsConfig's filter names to those used by `fluxLibrary`",
     )
 
 
 class FitBroadbandSEDTask(Task):
-    """Fit an observed SED with synthetic model SEDs
-    """
+    """Fit an observed SED with synthetic model SEDs"""
 
     ConfigClass = FitBroadbandSEDConfig
     _DefaultName = "fitBroadbandSED"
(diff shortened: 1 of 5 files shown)


python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -134,7 +144,9 @@ def run(self, pfsConfig):
 
         return pdfs
 
-    def getProbabilities(self, filterNames, bbFlux, bbFluxErr):
+    def getProbabilities(
+            self, filterNames: Sequence[str], bbFlux: Sequence[float], bbFluxErr: Sequence[float]) \
+            -> NDArray[numpy.float64]:
```
PaulPrice, line 149: Please look how `black` formats this.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -61,11 +70,12 @@ class FitBroadbandSEDTask(Task):
     ConfigClass = FitBroadbandSEDConfig
     _DefaultName = "fitBroadbandSED"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
```
PaulPrice, line 73: The `-> None` is implied if it's not there, so you can leave it off.

python/pfs/drp/stella/fitBroadbandSED.py
```diff
@@ -88,7 +98,7 @@ def runDataRef(self, dataRef):
         pfsConfig = dataRef.get("pfsConfig")
         return self.run(pfsConfig)
 
-    def run(self, pfsConfig):
+    def run(self, pfsConfig: PfsConfig) -> List[Union[NDArray[numpy.float64], None]]:
```
PaulPrice, line 101: I think this would read easier as `List[Optional[NDArray[np.float64]]]`. You can also use `List[NDArray[np.float64] | None]`.

PaulPrice (approved):
Nice.
I think your first commit combines choice of broadband flux with some flagging improvements that you may want to separate if you can.
I would encourage you to look at how `black` formats the files after you've marked them up with types; eventually we'll be adopting `black` to format everything but I've not gotten there yet.
Does `mypy` run clean on these files?

python/pfs/drp/stella/fitBroadbandSED.py
SogoMineo: To use the disjunction operator `|`, we need python-3.10. We are using python-3.8. The future introduction of the construction `A|None` is the reason why I use `Union[A, None]` instead of `Optional[A]`. I don't think `Optional` will survive long.

python/pfs/drp/stella/fitBroadbandSED.py
SogoMineo: I don't think so. If no return type is specified, `-> Any` is implied. What is worse, if none of the parameter types nor the return type is specified just as this `__init__()` function, then `mypy` does not  do type checking on the function body in my experience.

SogoMineo: By running mypy on this program, you can see how mypy skips a function whose signature is not specified.
```
def f():
    return 1 + "2"
```
```
$ mypy a.py
Success: no issues found in 1 source file
```

python/pfs/drp/stella/fitBroadbandSED.py
PaulPrice: Oh, sorry. I crossed wires with LSST, which has moved to python 3.10. We'll inherit that eventually, but not yet.

SogoMineo:
>Does mypy run clean on these files?
Not very. One problem is I don't know how to type `Struct`. I might replace it with `TypedDict` or `dataclasses`.

PaulPrice:
Yeah, `Struct` is a problem. There's a ticket to fix it on the LSST side.

SogoMineo:
`black` worked very well. The only modification I had to do manually was to join two short strings in a line:
```
"abc"
" def"
```
↓ black
```
"abc" " def"
```
↓ me
```
"abc def"
```

PR #306, opened 2023-01-24
PIPE2D-1147: estimateRadialVelocity: Change `findMethod` to "quadratic"

SogoMineo:
Previously, we found the peak position of the cross correlation function
by fitting a Gaussian to the c.c.f. Because exceptions were sometimes
raised by the optimizer, we change the peak-finding method to quadratic
interpolation. Because radial velocities found by the latter method
have large variance when `searchStep` is small, we change it from 5 km/s
to 10 km/s (This step should not be wider than the smallest possible
LSF). In addition, we lower S/N used in the unit test according to
true observed spectra.

### DIFF
--- a/python/pfs/drp/stella/estimateRadialVelocity.py
+++ b/python/pfs/drp/stella/estimateRadialVelocity.py
@@ -9,6 +9,72 @@ import scipy.optimize
 
 import math
 
+from typing import Sequence
+from typing import Any
+
+try:
+    from numpy.typing import NDArray
+except ImportError:
+    NDArray = Sequence
+
+
+__all__ = ["EstimateRadialVelocityConfig", "EstimateRadialVelocityTask"]
+
+
+class CrossCorrelationFunction:
+    r"""Cross correlation function:
+    ccf[k] = \sum_{i} flux[i] model(\lambda[i] * scale[k]).
+
+    Parameters
+    ----------
+    flux : `np.ndarray` of `float`
+        Shape ``(numWavelength,)``. The array ``flux`` in the equation above.
+    variance : `np.ndarray` of `float`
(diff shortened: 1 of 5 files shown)


python/pfs/drp/stella/estimateRadialVelocity.py
```diff
@@ -143,6 +144,49 @@ def run(self, spectrum: PfsFiberArray, modelSpectrum: PfsSimpleSpectrum) -> Stru
             # We have to think of the error...
             return Struct(velocity=velocity, error=np.nan, fail=fail, crossCorr=crossCorr)
 
+        if self.config.findMethod == "quadratic":
```
PaulPrice, line 147: I suggest moving the code under the {{if}} blocks into their own methods. This will make the code easier to read, and allow subclassing to change behaviour later.

PaulPrice (approved):
Only one minor comment.

RobertLuptonTheGood:
I'm confused by the statement "Because radial velocities found by the latter method have large variance when searchStep is small".  The error should be given by the curvature of the cross correlation function, which is independent of any step.

SogoMineo:
The error is not given by the curvature because a cross correlation function is not a probability distribution function. The error is given by the partial derivatives of `x(c_0, c_1, c_2)`, where `c_0`, `c_1`, and `c_2` are three samples of the cross correlation function and `x` is the peak position of the parabola passing through these samples. The samples `c_0`, `c_1`, and `c_2` are around the top of the peak. If `searchStep` is small, these samples are so close to each other that their differences will be smaller than noise: `|c_{i+1} - c{i}|^2 << Δc_{i+1}^2 + Δc_{i}^2`. Then, the peak position of the parabola gets noisy since the peak position is determined by their differences `c_1 - c_0` and `c_2 - c_1`.

PR #308, opened 2023-01-27
PIPE2D-1149: Make fitPfsFluxReference & fitFluxCal compatible with Gen3 butler

SogoMineo:
Added the elements necessary for `FitPfsFluxReferenceTask` and `FitFluxCalTask` to be compatible with Gen3 butler.

### DIFF
--- a/pipelines/science2.yaml
+++ b/pipelines/science2.yaml
@@ -1,0 +1,12 @@
+description: Full 2D-DRP science pipeline with fitPfsFluxReference and fitFluxCal.
+imports:
+  - $DRP_STELLA_DIR/pipelines/reduceExposure.yaml
+tasks:
+  fitPfsFluxReference:
+    class: pfs.drp.stella.fitPfsFluxReference.FitPfsFluxReferenceTask
+    config:
+      minBroadbandFluxes: 1
+  fitFluxCal:
+    class: pfs.drp.stella.fitFluxCal.FitFluxCalTask
+  coaddSpectra:
+    class: pfs.drp.stella.coaddSpectra.CoaddSpectraTask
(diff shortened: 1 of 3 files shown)


PaulPrice (approved):
Is it worth inheriting from the prototype flux calibration task instead of duplicating everything?

SogoMineo:
I think `*Connection` classes can inherit from, or be replaced by, the prototype one. However, I don't want `FitFluxCalTask` to directly inherit from `FluxCalibrateTask`, for some member functions have different parameters. If I am allowed to rewrite the prototype `FluxCalibrateTask` so that it will inherit from a base class, I can let `FitFluxCalTask` inherit from the same base class, too.

PaulPrice:
That's a good idea.

SogoMineo:
I tried to extract common parts of `FluxCalibrateTask` and `FitFluxCalTask`, and I found that their `run()` methods have different signatures from each other. `FluxCalibrateTask.run()` takes `Dict[int, PfsReference]` whereas `FitFluxCalTask.run()` takes `PfsFluxReference`. The former `run()` cannot adjust itself to the latter because of Gen2 interface. I don't think it worth to make a common base class that does not even have `run()`.

SogoMineo:
I delete `FitFluxCalConnections` and use `FluxCalibrateConnections` instead. I think this is reasonable because the input and output of `FitFluxCalTask` are the same as those of `FluxCalibrateConnections` in gen3.

SogoMineo:
Can I merge the current branch, or should I work on deleting more common lines?

PaulPrice:
Please go ahead an merge. Thanks!

PR #315, opened 2023-03-15
PIPE2D-1182: fitPfsFluxReference: More debug outputs

SogoMineo:
Added debug option (doWriteChisq) to output chi^2. With doWriteWhitenedFlux option, whitened model flux is also written together with observed flux.

I had to make somewhat many changes: To output degree of freedom with chi^2, I had to make a special class ChisqList. To reproduce the best-fit model as it was when it was compared to the observed flux, I had to make dopplerShift() function.

### DIFF
--- a/python/pfs/drp/stella/utils/math.py
+++ b/python/pfs/drp/stella/utils/math.py
@@ -1,8 +1,10 @@
+import dataclasses
 from types import SimpleNamespace
+from typing import Union
 import numpy as np
 from numpy.typing import ArrayLike
 
-__all__ = ("robustRms", "fitStraightLine")
+__all__ = ("robustRms", "fitStraightLine", "ChisqList")
 
 
 def robustRms(array: ArrayLike) -> float:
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/utils/math.py
```diff
@@ -49,8 +50,57 @@ def fitStraightLine(xx: np.ndarray, yy: np.ndarray) -> SimpleNamespace:
...
+    xySum = np.sum(dx * dy)
     xxSum = np.sum(dx**2)
-    slope = xySum/xxSum
-    intercept = yMean - slope*xMean
+    slope = xySum / xxSum
+    intercept = yMean - slope * xMean
     return SimpleNamespace(slope=slope, intercept=intercept, xMean=xMean, yMean=yMean)
+
+
+class ChisqList:
```
PaulPrice, line 60: Maybe use `dataclass`?

PaulPrice (approved):
Trivial comments only.

PR #320, opened 2023-05-11
PIPE2D-1208: Make fitPfsFluxReference even faster

SogoMineo:
This pull request contains three main changes:
1. Pack model templates into one uncompressed zip file.
2. Downsample model templates before continuum fitting.
With these two changes, fitPfsFluxReference in the integration test will run in less than 1 hour/visit/core with "full" model set.

3. Remove the filter name conversion table from FitBroadbandSEDTask.
This change will make the class simpler, though it will break compatibility to old fluxmodeldata packages. Users must install the new versions available at https://hscdata.mtk.nao.ac.jp/hsc_bin_dist/pfs/fluxmodeldata-ambre-20230428-small.tar.xz and https://hscdata.mtk.nao.ac.jp/hsc_bin_dist/pfs/fluxmodeldata-ambre-20230428-full.tar.xz .

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -22,11 +22,10 @@ from pfs.datamodel.observations import Observations
 from pfs.datamodel.pfsConfig import FiberStatus, PfsConfig, TargetType
 from pfs.datamodel.pfsFiberArray import PfsFiberArray
 from pfs.datamodel.pfsFluxReference import PfsFluxReference
-from pfs.datamodel.pfsSimpleSpectrum import PfsSimpleSpectrum
 from pfs.drp.stella.fluxModelInterpolator import FluxModelInterpolator
 from pfs.drp.stella import ReferenceLine, ReferenceLineSet, ReferenceLineStatus
 from pfs.drp.stella import SpectrumSet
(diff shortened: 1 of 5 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -1026,6 +1056,42 @@ def getBadMask(self) -> List[str]:
...
+            Downsampled model.
+        """
+        if self.config.modelOversamplingFactor <= 0:
+            # Downsampling is disabled
+            return model
+
+        minWavelength = model.wavelength[0]
+        maxWavelength = model.wavelength[-1]
+        dlambdaModel = np.max(model.wavelength[1:] - model.wavelength[:-1])
+        dlambdaObs = np.min(obsSpectrum.wavelength[1:] - obsSpectrum.wavelength[:-1])
```
PaulPrice, line 1082: Might want to use `pfs.drp.stella.interpolate.calculateDispersion`?

PaulPrice (approved):
Trivial comments.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: `calculateDispersion` returns an array, but I want scalars here.

PR #325, opened 2023-06-05
PIPE2D-1231: fluxModelInterpolator: Use PCA to reduce memory usage

SogoMineo:
fluxModelInterpolator.py after this commit requires fluxmodeldata-ambre-20230602.

### DIFF
--- a/python/pfs/drp/stella/fluxModelInterpolator.py
+++ b/python/pfs/drp/stella/fluxModelInterpolator.py
@@ -4,20 +4,172 @@ from pfs.datamodel.wavelengthArray import WavelengthArray
 from pfs.drp.stella.datamodel.pfsFiberArray import PfsSimpleSpectrum
 
 import astropy.io.fits
+from deprecated import deprecated
 import numpy as np
 import scipy.interpolate
 
+import json
 import os
 import pickle
 
 __all__ = ("FluxModelInterpolator",)
 
 
+class PCACompositor:
+    """Given ``mean`` and basis vectors, compute ``mean`` plus
+    linear combinations of the basis vectors.
+
+    Parameters
+    ----------
+    basis : `np.ndarray`
+        Basis vectors. Shape (M, N), M being the number of basis vectors
+        and N being the dimension of the vectors.
+
+    mean : `np.ndarray`
+        Mean. Shape (N,). This vector is added to linear combinations
+        of basis vectors.
+
+    header : `astropy.io.fits.Header`
+        FITS header that contains WCS cards
+        for ``mean`` and for vectors output from ``__call__()``.
+    """
+
+    def __init__(self, basis: np.ndarray, mean: np.ndarray, header: astropy.io.fits.Header) -> None:
+        self.basis = np.asarray(basis)
+        self.mean = np.asarray(mean)
+        self.header = header
+
+    def __call__(self, coeff: np.ndarray) -> np.ndarray:
+        r"""Compute a linear combination (+ ``self.mean``).
+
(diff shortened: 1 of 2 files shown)


python/pfs/drp/stella/fluxModelInterpolator.py
```diff
@@ -79,38 +155,25 @@ def fromFluxModelData(cls, path: str) -> "FluxModelInterpolator":
...
             The model.
         """
-        filePath = os.path.join(path, "interpolator.pickle")
-        if not os.path.exists(filePath):
-            raise RuntimeError(f"'{filePath}' not found. Run `makeFluxModelInterpolator.py` to generate it.")
-        return cls.fromPickle(filePath)
+        interpPath = os.path.join(path, "interpolator.pickle")
+        if not os.path.exists(interpPath):
+            raise RuntimeError(
+                f"'{interpPath}' not found. Run `makeFluxModelInterpolator.py` to generate it."
```
PaulPrice, line 161: Do I have to run `makeFluxModelInterpolator.py` in addition to the `install.py` script? It should be part of the installation process.

python/pfs/drp/stella/makeFluxModelInterpolator.py
```diff
@@ -22,26 +21,6 @@ def makeFluxModelInterpolator(fluxmodeldataPath):
...
-    param = inputParameters[0]
-    spectrum = modelSet.getSpectrum(param["teff"], param["logg"], param["m"], param["alpha"])
-    fluxList = np.empty(shape=(len(inputParameters), len(spectrum)), dtype=float)
-    wcs = spectrum.wavelength.toFitsHeader()
-
-    # Store fluxes of each model into a single array
-    for i, param in enumerate(inputParameters):
-        spectrum = modelSet.getSpectrum(param["teff"], param["logg"], param["m"], param["alpha"])
-        fluxList[i, :] = spectrum.flux
-
```
PaulPrice, line 44: There have been some significant changes to the format of the fluxmodeldata package, some coming in rapid succession. I think we need to implement some sort of versioning to ensure compatibility between the fluxmodeldata and the code, and to allow the user to immediately see the problem in the case of a discrepancy.

python/pfs/drp/stella/fluxModelInterpolator.py
SogoMineo: `makeFluxModelInterpolator.py` was necessary in the past because it took very long time to make it, but now it takes only a few seconds to make the interpolator. I think I can remove `makeFluxModelInterpolator.py`. The installer must not make `interpolator.pickle` because I cannot control the version of python that users use on the installation.

python/pfs/drp/stella/fluxModelInterpolator.py
```diff
@@ -13,6 +13,81 @@
...
+        Returns
+        -------
+        sum : `np.ndarray`
+            Shape (..., N), N being the dimension of a basis vector.
+        """
+        coeff = np.asarray(coeff)
+        lenShape = len(coeff.shape)
+        if lenShape > 1:
+            ones = (1,) * (lenShape - 1)
+            mean = self.mean.reshape(ones + self.mean.shape)
```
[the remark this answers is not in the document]
SogoMineo: I don't think
```
addedAxes = tuple(range(lenShape - 1))
mean = np.expand_dims(mean, axis=addedAxes)
```
is more readable than what I wrote:
```
ones = (1,) * (lenShape - 1)
mean = self.mean.reshape(ones + self.mean.shape)
```
I like the latter because it is less cryptic.

PR #327, opened 2023-06-20
PIPE2D-1238: fitPfsFluxReference: Fit models using Powell's method

SogoMineo:
Previously, fitPfsFluxReference fitted models to spectra by evaluating
chi^2 for every precomputed model template. After this change,
fitPfsFluxReference uses Powell's method (by default) to fit models.

With this change, fitPfsFluxReference.py runs 3x faster (if the number
of FLUXSTD in a visit is 100). The quality of the output seems slightly
better nonetheless for the possibility of falling into a local minimum,
to my surprise. It may be that the sub-grid peak finder in the previous
method was bad, or that LSF approximation for speedup was problematic,
though these are mere guesses.

The previous method can be restored with
`config.minimizationMethod='brute-force'`.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -40,6 +40,9 @@ from pfs.drp.stella.utils import debugging
 
 from astropy import constants as const
 import numpy as np
+import scipy.interpolate
+import scipy.optimize
+import scipy.spatial
 
 import copy
 import dataclasses
@@ -154,13 +157,32 @@ class FitPfsFluxReferenceConfig(PipelineTaskConfig, pipelineConnections=FitPfsFl
         default="psf",
         optional=False,
     )
+    minimizationMethod = Field(
+        dtype=str,
+        default="Powell",
+        doc="Method with which to find the minimum of the objective function of spectrum fitting."
+        " Valid values are 'brute-force' or any valid arguments for `scipy.optimize.minimize()`",
+    )
     priorCutoff = Field(
         dtype=float,
         default=0.01,
-        doc="Cut-off prior probability relative to max(prior)."
+        doc="(Valid if minimizationMethod='brute-force') Cut-off prior probability relative to max(prior)."
         " In comparing a flux model to an observed flux,"
         " the model is immediately rejected if its prior probability"
         " is less than `priorCutoff*max(prior)`.",
+    )
+    paramScale = ListField(
+        dtype=float,
+        default=[1e-3, 1, 1, 1],
+        doc="(Valid if minimizationMethod!='brute-force')"
+        " Values by which to multiply (teff, logg, m, alpha)"
+        " to adjust them to be as large as each other.",
+    )
+    paramOrder = ListField(
+        dtype=int,
+        default=[2, 3, 0, 1],
+        doc="(Valid if minimizationMethod!='brute-force')"
+        " Order of (teff (0), logg (1), m (2), alpha (3)) in the parameter list of the objective function.",
     )
     modelOversamplingFactor = Field(
         dtype=float,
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -154,10 +157,16 @@ class FitPfsFluxReferenceConfig(PipelineTaskConfig, pipelineConnections=FitPfsFl
         default="psf",
         optional=False,
     )
+    minimizationMethod = Field(
```
PaulPrice, line 160: Use `ChoiceField` to limit the possible values?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -574,6 +583,187 @@ def fitModelsToSpectra(
...
+              - dof : `int`
+                degree of freedom of the chi square.
+        """
+        if self.config.minimizationMethod.lower() == "brute-force":
+            return self.fitModelsToSpectraBruteForce(
+                pfsConfig, obsSpectra, pfsMergedLsf, radialVelocities, priorPdfs
+            )
+
+        # Value by which to multiply T_eff to adjust it to be as large as the other parameters.
+        teffScale = 1e-3
```
PaulPrice, line 636: Make this a config parameter?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -574,6 +583,187 @@ def fitModelsToSpectra(
...
+                True if optimization has succeeded.
+
+            The following members exist if self.config.minimizationMethod != 'brute-force'
+
+              - chi2 : `float`
+                chi square.
+              - dof : `int`
+                degree of freedom of the chi square.
+        """
+        if self.config.minimizationMethod.lower() == "brute-force":
```
PaulPrice, line 630: I would have done this with an inherit-and-override approach, but your approach is easier to change on the fly.
I think it would be best to have the selection done outside the method that does the work, and have the method that does the work concentrate solely on the work to be done.

PaulPrice (approved):
This is a great move in the direction of faster fits. I have some questions/suggestions, but do what you feel is right.

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -721,7 +710,37 @@ def fitModelsToSpectra(
...
+        if self.debugInfo.doWriteChisq:
+            debugging.writeExtraData(
+                f"fitPfsFluxReference-output/chisq-{obsSpectra.filename}.pickle",
+                fiberId=pfsConfig.fiberId,
+                chisq=chisqLists,
+            )
+
+        # Posterior PDF
+        pdfs: List[Union[NDArray[np.float64], None]] = []
+        for prior, chisq in zip(priorPdfs, chisqLists):
```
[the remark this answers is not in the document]
SogoMineo: The problem is that the double loop is not `for each observed spectrum; for each model; do...` but `for each model; for each observed spectrum; do...`. The double loop must be in this order because it takes more time to prepare a model than to prepare an observed spectrum for comparison. Once the brute-force method is gone, I think it better for there to be a Task that fit a single fiber, for it would pave the way to parallelization.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: > an inherit-and-override approach
It was my first plan, but I thought they would have to use retarget to switch the tasks with a configuration file. I am not familiar with it.
> have the selection done outside the method
Whether the selection is done outside or not, I have to test `minimizationMethod` against the value "brute-force" just before passing it to scipy. If `minimizationMethod` turns out to be "brute-force" in the test, I have two choices: raise an exception or jump to the special method. I think the second choise is the better. Then, the selection outside is redundant.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I don't want this parameter to be restricted to "brute-force" or "Powell", but I want to allow any valid value that can be passed to scipy.optimize.minimize. I don't use ChoiceField here because scipy.optimize.minimize may be extended in future.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I decide to use FluxModelInterpolator's `teffScale`, `loggScale`, ... here instead of using the immediate value.

PR #331, opened 2023-07-18
PIPE2D-1259: fitContinuum.py: Make binData() faster

SogoMineo:
binData() is one of the hot spots during the execution of
fitPfsFluxReference.py. It can be made faster by removing the for-loop.

Binning results may slightly change due to the difference regarding
the treatment of the case when `numBins` does not divide `len(xx)`.


PaulPrice (approved):
Awesome, thanks!

PR #332, opened 2023-07-24
PIPE2D-1252: reduceExposure: Correct LSF widths to Gaussian sigmas.

SogoMineo:
The document of ReduceExposureConfig.gaussianLsfWidth says its elements are Gaussian sigmas, but the values actually written there were FWHM.


SogoMineo:
Rebased to the current master branch.

PR #334, opened 2023-09-06
PIPE2D-1237: fitPfsFluxReference: Discard FLUXSTD when its param is out of range

SogoMineo:
If the guessed parameter of a spectrum is out of the parameter domain
of the models we have prepared, we should discard the spectrum.

### DIFF
--- a/pipelines/science.yaml
+++ b/pipelines/science.yaml
@@ -6,6 +6,9 @@ tasks:
     class: pfs.drp.stella.fitPfsFluxReference.FitPfsFluxReferenceTask
     config:
       minBroadbandFluxes: 1
+      minTeff: 0
+      maxTeff: 9999
+      paramMargin: [0, 0, 0, 0]
   fitFluxCal:
     class: pfs.drp.stella.fitFluxCal.FitFluxCalTask
   coaddSpectra:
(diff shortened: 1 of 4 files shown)


pipelines/science.yaml
```diff
@@ -6,6 +6,9 @@ tasks:
     class: pfs.drp.stella.fitPfsFluxReference.FitPfsFluxReferenceTask
     config:
       minBroadbandFluxes: 1
+      minTeff: 0
+      maxTeff: 9999
+      paramMargin: [0, 0, 0, 0]
```
PaulPrice, line 11: Why are the integration test flux standards being rejected? Is there something wrong with them, or is it something wrong with the code?

PaulPrice (approved):
This is terrific, but I'm concerned that the integration test requires special treatment.

pipelines/science.yaml
SogoMineo: "gen2" test succeeds even if I replace fluxCalibrate.py with pfsFluxReference.py. Effective temperature of the three FLUXSTD are estimated to be ~7000K. But "gen3" test fails with the default configurations because effective temperature of the three FLUXSTD are 5500K, 8000K, and nan K. All the three FLUXSTD are rejected.
Sha1 hashes of `pfsArm` differ between gen2 and gen3. I have not investigated further.

PaulPrice:
OK, so this is something that we can put on the Gen3 pile. Fair enough.

SogoMineo:
I further tried to understand why gen3 integration test's results are different from gen2 integration test's, but all I found was that all files, including calibs, are different between gen2 and gen3, except for input files and curated calibs. Let me merge this branch as-is.

PR #335, opened 2023-09-25
PIPE2D-1301: reduceProfiles: update following reduceExposure API change

SogoMineo:
reduceExposure was previously updated to operate on a single exposure at a time rather than lists. It seems there weren't corresponding changes in reduceProfiles.

### DIFF
--- a/src/FiberTraceSet.cc
+++ b/src/FiberTraceSet.cc
@@ -181,7 +181,8 @@ SpectrumSet FiberTraceSet<ImageT, MaskT, VarianceT>::extractSpectra(
                 if (modelValue > minFracMask) {
                     maskResult[ii] |= maskValue;
                 }
-                if ((maskValue & badBitMask) || !std::isfinite(imageValue) || !std::isfinite(varianceValue)) {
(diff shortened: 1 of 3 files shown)


src/FiberTraceSet.cc
```diff
@@ -181,7 +181,8 @@ SpectrumSet FiberTraceSet<ImageT, MaskT, VarianceT>::extractSpectra(
                 if (modelValue > minFracMask) {
                     maskResult[ii] |= maskValue;
                 }
-                if ((maskValue & badBitMask) || !std::isfinite(imageValue) || !std::isfinite(varianceValue)) {
+                if ((maskValue & badBitMask) || !std::isfinite(imageValue) || !std::isfinite(varianceValue) ||
+                    varianceValue == 0) {
```
SogoMineo, line 185: Is `varianceValue` never negative?

SogoMineo:
The changes look good to me, but I don't have the permission to approve the changes because I made this PR.

src/FiberTraceSet.cc
PaulPrice: It should never be zero either, so probably worth making this `<=`.

PR #338, opened 2023-10-25
Handle failed aperture corrections by setting fluxes to NaN

RobertLuptonTheGood:
A cleverer way would be to build an apCorr object full of NaNs, but I don't know how to do that.

N.b.  Other recovery strategies are possible, but this is simple. We need to label the line fluxes as bad, but it would be nice to keep the raw fluxes too


PaulPrice (approved):
Great, thanks!

PR #343, opened 2024-01-31
PIPE2D-1362: fitFluxCal should correct seeing variation

SogoMineo:
The type of fluxCal is no longer ConstantFocalPlaneFunction but another
subclass of FocalPlaneFunction. It consists of (1) an average flux
calibration vector up to scales and (2) the exponential of a polynomial
of (x, y, lambda), which is expected to be a gently varying function
of fiber position and wavelength, representing scale (i.e. overall
height of a flux calibration vector) that vary according to fiber
position in the field of view, and seeing variation according to
wavelength.


PaulPrice (approved):
Awesome work.

PR #346, opened 2024-03-27
Tickets/pipe2 d 1409

RobertLuptonTheGood:


yukimoritani:
I ran the notebook and looks good. If I can make a request, could you set y-axis label as integer in _throughputPerSpectrograph_
(e.g., plt.yticks([1,2,3,4]) )?

cf.  The notebook for this modification was pushed under tickets/PIPE2D-1408

PR #352, opened 2024-05-20
PIPE2D-1446: fitPfsFluxReference: Use diff. w.l.ranges for mid- and low-res spectra

SogoMineo:
In fitting a flux reference, we use different wavelength ranges for mid-res spectra and low-res spectra. Whether spectra are mid-res or low-res is determined not by pfsConfig but by the spectra themselves because pfsConfig's information is unreliable: pfsConfig.arms may be "brn" when it should be "bmn". The reason for this is that humans may disobey pfsDesign in actual observation.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -116,7 +116,7 @@ class FitPfsFluxReferenceConfig(PipelineTaskConfig, pipelineConnections=FitPfsFl
     )
     minWavelength = Field(
         dtype=float,
-        default=600.0,
+        default=480.0,
         doc="min of the wavelength range in which observation spectra are compared to models.",
     )
     maxWavelength = Field(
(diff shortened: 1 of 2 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -1837,6 +1856,75 @@ def marginalizePdf(
...
+    ----------
+    pfsMerged : `PfsMerged`
+        Merged spectra from exposure.
+
+    Returns
+    -------
+    midres : `bool`
+        True if a part of spectra in ``pfsMerged`` are from mid-resolution arm.
+    """
+    return any("m" in arms for arms in getExistentArms(pfsMerged).values())
```
PaulPrice, line 1876: Why not use `"m" in pfsMerged.identity.arm`?

PaulPrice (approved):
I think the determination of whether `m` is involved can be greatly simplified.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: Thanks for the information. I will use `pfsMerged.identity.arm` here, and move `getExistentArms()` to `tickets/PIPE2D-1427`

PR #356, opened 2024-06-18
Treat all NaNs as equivalent in magnitude arrays

RobertLuptonTheGood:


python/pfs/drp/stella/coaddSpectra.py
```diff
@@ -33,6 +33,21 @@
...
+        """Construct a set from iterable with at most one NaN"""
+        sawNaN = False
+
+        for x in iterable:
+            if np.isnan(x):
+                if not sawNaN:
+                    self.add(x)
+                    sawNaN = True
+            else:
+                self.add(x)
```
PaulPrice, line 48: I expect this does what we need, but if we were to use this more widely it would need a check in the `add` method as well.

PaulPrice (approved):

PR #357, opened 2024-07-01
Don't apply screen corrections to IIS fibres

RobertLuptonTheGood:
The code handles mixed IIS/PFI fibres, but we also check for all-ENGINEERING in order to generate better log messages


PaulPrice (approved):
Good catch!

PR #364, opened 2024-11-20
fitFluxCal: Apply median filter to spectra before integrating them

SogoMineo:
We apply a median filter to the spectra of flux standards before integrating them (to compute broadband fluxes) so that outlier pixels in low S/N spectra will not corrupt the integrals.


wtgee (approved):
LGTM

PR #368, opened 2025-01-20
PIPE2D-1518: Fix flux normalization variation in flux calibration

SogoMineo:
This is a collection of several improvements to flux calibration.

### DIFF
--- a/python/pfs/drp/stella/fitFluxCal.py
+++ b/python/pfs/drp/stella/fitFluxCal.py
@@ -1139,11 +1139,11 @@ class FitFluxCalibFocalPlaneFunctionConfig(FitFocalPlaneConfig):
     input parameters, but those can be determined from the data.
     """
 
-    polyOrder = Field(dtype=int, default=3, doc="Polynomial order")
+    polyOrder = Field(dtype=int, default=5, doc="Polynomial order")
     polyWavelengthDependent = Field(
         dtype=bool,
-        default=False,
-        doc="Whether the polynomial is wavelength-dependent. ('this option=True' is deprecated)",
+        default=True,
+        doc="Whether the polynomial is wavelength-dependent.",
     )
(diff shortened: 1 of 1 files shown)


python/pfs/drp/stella/fitFluxCal.py
```diff
@@ -295,6 +295,8 @@ class BroadbandFluxChi2:
         Mask planes for bad pixels.
     softenBroadbandFluxErr: `float`
         Soften broadband flux errors: err**2 -> err**2 + (soften*flux)**2
+    unusedBroadbandFilters: `list` [`str`]
```
PaulPrice, line 298: Maybe "ignored" is better than "unused"?

python/pfs/drp/stella/fitFluxCal.py
```diff
@@ -342,7 +345,11 @@ def __init__(
         self.arms: dict[int, str] = getExistentArms(pfsMerged)
 
         self.bbFlux: dict[int, list[tuple[float, float, str]]] = {
-            fiberId: list(zip(bbFlux, bbFluxErr, filterNames))
+            fiberId: [
+                (flux, err, filter)
+                for flux, err, filter in zip(bbFlux, bbFluxErr, filterNames)
+                if filter not in unusedBroadbandFilters
```
PaulPrice, line 351: Probably should cast `unusedBroadbandFilters` to a `set` for efficiency.

PaulPrice (approved):

python/pfs/drp/stella/fitFluxCal.py
SogoMineo: I was not sure whether or not `set` outperforms `list` when the number of elements is only a few, but experiment shows `set` is indeed faster when there are only two elements.

PR #369, opened 2025-01-20
PIPE2D-1592: fitPfsFluxReference: Don't use low-SNR flux standards

SogoMineo:
Flux standards whose S/N are extremely low should not be used even if their flags are clear.

### DIFF
--- a/python/pfs/drp/stella/fitPfsFluxReference.py
+++ b/python/pfs/drp/stella/fitPfsFluxReference.py
@@ -46,6 +46,7 @@ import scipy.spatial
 import copy
 import dataclasses
 import math
+import warnings
 
 from typing import Literal, overload
 from collections.abc import Generator, Mapping, Sequence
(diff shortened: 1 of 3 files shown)


python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -1996,6 +2021,42 @@ def removeBadFluxes(
...
+    pfsMerged : `PfsMerged`
+        PfsMerged from which bad spectra have been removed.
+    """
+    left, right = cutoffSNRRange
+    good = np.zeros(shape=(len(pfsMerged),), dtype=bool)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        for i in range(len(pfsMerged)):
+            sampleIndex = (left <= pfsMerged.wavelength[i]) & (pfsMerged.wavelength[i] <= right)
+            snr = pfsMerged.flux[i, sampleIndex] / np.sqrt(pfsMerged.covar[i, 0, sampleIndex])
```
PaulPrice, line 2054: Maybe `pfsMerged.variance[i, sampleIndex]`?

python/pfs/drp/stella/fitPfsFluxReference.py
```diff
@@ -402,6 +419,14 @@ def selectPfsConfig(pfsConfig: PfsConfig, flagName: str, isGood: Sequence[bool])
             return goodConfig
 
         pfsConfig = selectPfsConfig(pfsConfig, "ABSENT_FIBER", np.isin(pfsConfig.fiberId, pfsMerged.fiberId))
+
+        pfsMerged = removeBadSpectra(
+            pfsMerged,
+            self.config.cutoffSNR,
+            (self.config.cutoffSNRRangeLeft, self.config.cutoffSNRRangeRight),
+        )
```
PaulPrice, line 427: Need to check the case that `len(pfsMerged) = 0` after calling `removeBadSpectra`.

PaulPrice (approved):

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I don't think so. That condition is checked tens of lines below, before pfsMerged is actually used.

python/pfs/drp/stella/fitPfsFluxReference.py
SogoMineo: I am adding another commit to this PR, in which I replace all occurrences of "covar" with "variance" in the entire flux calibration code.

PR #370, opened 2025-02-11
PIPE2D-1645: Don't assume that darks are on-sky exposures

RobertLuptonTheGood:


RobertLuptonTheGood:
I didn't like looking at the header; you're right, I could use `exp.info.getVisitInfo().observationType`.  Coming up

RobertLuptonTheGood:
Except that there's no `exposure` in this function.   Hmm

PR #371, opened 2025-02-24
Fix lsst Gen3/v28 issues - tickets/PIPE2D-1658

wtgee:
* Updates for butler lookups and exceptions.


wtgee:
I've tested the imports of functions and they work. The `lines` lookup and the `DataSetNotFound` error are both within the animation function, which doesn't actually get much use. We could make another ticket to make that work if we want but this ticket is just for getting the imports of other functions to work.

PaulPrice (approved):

PR #372, opened 2025-03-12
PIPE2D-1643: Improve resampling for medium resolution

PaulPrice:


python/pfs/drp/stella/lsf.py
```diff
@@ -1,26 +1,40 @@
 from abc import ABC, abstractmethod
 from functools import partial
+from typing import TYPE_CHECKING, Callable, Iterable, Iterator
```
SogoMineo, line 3: `Callable`, `Iterable`, and `Iterator` in `typing` module have been deprecated since python-3.9. They should be imported from `collections.abc`

SogoMineo (approved):
Looks good to me. Also see my comment in JIRA.

PR #377, opened 2025-06-21
np.NaN -> np.nan for numpy 2.x

CraigLoomis:


CraigLoomis:
Many/most changes already made on master.

PR #384, opened 2025-09-12
PIPE2D-1764: Handle processing unilluminated ENGINEERING fibres

RobertLuptonTheGood:


PaulPrice (approved):
I hate backslashes, so would prefer you change that, but otherwise good.

PR #390, opened 2026-04-22
feat: add plotDir support and debug_detectormap.py script

wtgee:
Add _showOrSavePlot() helper to FitDistortedDetectorMapTask that saves all open matplotlib figures to <plotDir>/<name>[_N].png when debugInfo.plotDir is set, or falls back to plt.show() for interactive use. Replace all 7 bare plt.show() calls in the task with named _showOrSavePlot() calls.

Add bin.src/debug_detectormap.py: a ready-to-use lsstDebug override that enables all diagnostic plot flags. The output directory is controlled via DETECTORMAP_PLOT_DIR and the set of enabled flags via DETECTORMAP_DEBUG_FLAGS.


PaulPrice:
`debug_detectormap.py` is not an executable. Shall we put it in a new `debug` directory?

wtgee:
> `debug_detectormap.py` is not an executable. Shall we put it in a new `debug` directory?

Yes, good call, thanks. I'm going to test this out and make sure it works first then will clean the PR.  Would be nice to start utilizing the existing plots if easy enough.

PaulPrice:
👍 
That's a good find -- I didn't know about that functionality.

